PR #3, opened 2018-09-22
Add Apache license text

pbugnion:
Having an explicit license:
- allows people to use the package, since they know exactly under what conditions they are allowed to use it.
- limits the liability for developers.

The [Apache license](https://en.wikipedia.org/wiki/Apache_License) is a permissive license inasmuch as it allows using and re-distributing the code for commercial and non-commercial use. It's generally considered a good model for code released by organisations. It's very close to the MIT license that is generally used in the Pydata stack. It's also what we use for most open source packages we distribute.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,0 +1,201 @@
+                              Apache License
+                        Version 2.0, January 2004
+                    http://www.apache.org/licenses/
+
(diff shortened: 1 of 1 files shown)


pbugnion:
Great, thanks!

PR #10, opened 2018-09-28
Prototype documentation

pbugnion:
This PR adds the initial framework for documentation.

For demo purposes, this is deployed on [SherlockML](https://dash-bootstrap-components-documentation.app.asi.sherlockml.net/) for now. Because copying is easy and creativity is hard, the style of the documentation comes mostly from the [reactstrap](https://reactstrap.github.io/) docs, with bits taken from [photon](http://photonkit.com/).

This is very much a prototype, but I think the initial framework is sufficiently complete that we could consider merging it. What's left to do on documentation after this:
- validate deployment on, e.g. an EC2 machine or a Digital Ocean droplet.
- add more components
- better tutorial documentation on structuring layouts
- examples


pbugnion:
This is blocked until we have navs, which I'm currently working on.

pbugnion:
@tcbegley I think this is now ready for review.

tcbegley:
This looks absolutely brilliant, thanks for getting it set up @pbugnion!

My only comment is on the syntax highlightingon the components page, it seems decorators are getting highlighted as though they were comments:

image

Looks like it's the `dcc.SyntaxHighlighter` that's doing it, I don't know how configurable that is?

pbugnion:
That's a good catch -- actually, the style of the code in the landing page (where I've just used prismjs directly) is different to the style that dash-core-components uses for its `SyntaxHighlighter`. PrismJS has different themes and I must have just picked the wrong one. I'll look into what dash-core-components uses and check to see why their decorators look like comments.

pbugnion:
I think this is ready for another look. 

By passing `useInlineStyles=False` to `SyntaxHighlighter`, the component just wraps code in spans with classes that denote the type of token. We can then use any [HighlightJS](https://highlightjs.org/) stylesheet for the styles.

I settled for Monokai, but changing `hljs-meta` to the monokai cyan so decorators don't look like comments, but happy to change the style if you can think of a better stylesheet.

tcbegley:
This is looking really great!

What do you think is the best way to proceed now? Now that we have something in place I'm happy to start contributing content.

pbugnion:
Thanks!

I think the next step would be to merge this PR. Then:
1. Whenever we add components, we should add a documentation entry
2. Gradually backfill documentation on existing components (mostly by copying the existing examples)
3. Identify material that needs a to be explained through tutorials, rather than just as an entry in the components page. I'd imagine things related to layout etc. would be good candidates for this.

What do you think?

tcbegley:
I think that sounds great!

PR #13, opened 2018-09-29
Jumbotron dropdown

tcbegley:
This PR adds a dropdown component and a jumbotron component

@pbugnion I'm a bit unsure about using "Dropdown" as it clashes with the Dropdown component in dash core components. Should we try to find a better name? Menu, or DropdownMenu for example?

### DIFF
--- a/demo/Demo.js
+++ b/demo/Demo.js
@@ -21,6 +21,7 @@ import {
   Dropdown,
   DropdownItem,
   Fade,
+  Jumbotron,
   ListGroup,
   ListGroupItem,
   ListGroupItemHeading,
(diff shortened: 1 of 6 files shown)


pbugnion:
I think dropdown has a much more precise meaning than menu, so `Dropdown` is a better name.

I don't have strong preferences between `Dropdown` and `DropdownMenu`, except that I *don't* think shadowing dash core components is particularly bad (since we consistently use the `dbc` namespace in examples).

tcbegley:
Ok cool, I'll leave it as it is for now, thanks for your input!

PR #14, opened 2018-09-30
DashLink: set default behaviour based on href

pbugnion:
Prior to this PR, the default behaviour for `DashLink` depended on the `refresh` prop and the `dashLink` prop. As far as I can tell, both of these did the same thing:

- if `refresh` was true or `dashLink` was false, this forced a page reload at href.
- if `refresh` was false or `dashLink` was true, a new href was pushed onto the browser history and an event was dispatched, presumably to be caught by the backend.

This PR changes the behaviour by:
- removing `refresh` and `dashLink` completely.
- having a single `external_link` prop that allows the user to control whether the link should trigger a full page reload.
- if `external_link` is not set, choosing the default behaviour based on `href`. If `href` is an external location, prefer to force a reload. If `href` is internal, fall back onto Dash for managing the internal state.

Besides this, this change also introduces unit tests on the client. The behaviour of `DashLink` is pretty fiddly, so I thought it'd be usefu. The tests uses Jest and Enzyme, which are the de-facto standard in the React world. If we think the unit tests make sense, I'm happy to set up Travis so we can re-run them on regressions.

Outstanding questions:
- At the moment, `DashLink` is exposed. Is the plan to actually expose it to users? If not, we could consider keeping it private? Not sure how we would stop Dash from creating a Python class for it, but I expect we can figure it out.
- `DashLink` isn't a great name. I'd probably prefer to fall back onto just `Link`, specially if it's just going to be used internally.

----

I have used this test script to test the behaviour:

```py
import dash

app = dash.Dash(
    __name__, 
    external_stylesheets=['https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/css/bootstrap.min.css']
)

app.layout = dbc.Container([
    dcc.Location(id='url', refresh=False),
    html.H2('CardLink demonstration'),
    dbc.CardDeck([
        dbc.Card([
            dbc.CardBody([
                # Use internal redirection -- does not trigger page reload
                dbc.CardLink('internal', href='/internal'),
                
                # Use external redirection -- triggers a page reload
                dbc.CardLink('external', href='https://www.asidatascience.com'),
                
                # Redirect internally, but force a page reload. Useful for
                # fetching internal pages not under the control of the current
                # browser client.
                dbc.CardLink('external-explicit', href='/external', external_link=True),
                
                # Redirect internally without forcing a page reload.
                # This is questionably useful. The href needs
                # to be the same host as the current host.
                # Does not trigger a page load.
                dbc.CardLink('internal-explicit', id='internal-explicit', external_link=False)
            ])
        ])
    ]),
    html.P(id='url-value')
])


@app.callback(
    Output('internal-explicit', 'href'),
    [Input('url', 'href')]
)
def set_internal_href(current_href):
    return f'{current_href}/internal'


@app.callback(
    Output('url-value', 'children'),
    [Input('url', 'pathname')]
)
def on_url_change(pathname):
    return ['current path: ', pathname]


if __name__ == '__main__':
```

### DIFF
--- a/demo/Demo.js
+++ b/demo/Demo.js
@@ -12,6 +12,7 @@ import {
   CardHeader,
   CardImg,
   CardImgOverlay,
+  CardLink,
   CardSubtitle,
   CardText,
   CardTitle,
@@ -201,7 +202,7 @@ class Demo extends Component {
       <h2>Cards</h2>
 
       <Row>
-        <Col md={6}>
+        <Col md={3}>
           <Card color="dark" outline={true}>
             <CardHeader>Header</CardHeader>
             <CardBody>
@@ -225,6 +226,16 @@ class Demo extends Component {
               <CardTitle>Here's another card</CardTitle>
               <CardText>With some text, and a button</CardText>
               <Button color="success">Click here</Button>
+            </CardBody>
+          </Card>
+        </Col>
+        <Col md={3}>
+          <Card>
+            <CardBody>
+              <CardTitle>Here's yet another card</CardTitle>
+              <CardText>With some text, and some links</CardText>
+              <CardLink href="https://www.asidatascience.com">External</CardLink>
+              <CardLink href="/example">Internal</CardLink>
             </CardBody>
           </Card>
         </Col>
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,40 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
+      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
+      "dev": true,
+      "requires": {
+        "@babel/highlight": "^7.0.0"
+      }
+    },
+    "@babel/highlight": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.0.0.tgz",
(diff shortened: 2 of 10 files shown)


tcbegley:
Thanks for this @pbugnion!

I think the new interface is really nice, certainly a lot cleaner than it was before.

Unit tests would definitely be useful, if you are happy to set them up that would be great.

I agree `DashLink` should not be exposed. Looking at the `extract-metadata` script in `package.json` I think maybe all we need to do is move it out of `src/components` and into something like `src/base`? I just tried this quickly and it seems to work, but we should probably check it more carefully.

I am also happy to fall back to `Link` as opposed to the slightly clunky `DashLink`.

pbugnion:
Great, thanks!

I've moved `DashLink` to `src/private`, and renamed it to `Link`. I thought `private` was more explicit than `base`. We might want to have a `base` in `components` later on.

The Travis setup is in a separate PR (#15).

pbugnion:
There's actually one more outstanding question. In this, I used `snake_case` for `external_link`, since it was exposed in Python. Somewhat confusingly, Dash seems to use a mix of snake case (e.g. `n_clicks`, `end_date_placeholder_text`) and camel case (e.g. `className`, `labelStyle`) for attributes.

Any preferences? We should probably choose one and stick to it. I personally think it makes more sense to use snake case, since the interface is in Python, but happy to be convinced either way.

pbugnion:
I noticed that we could just use the `tag` prop to replace an anchor with our own link in the higher level wrappers of links (like `CardLink`, `NavLink` and `NavbarBrand`). This is nice because it means we don't have to re-implement reactstrap logic for injecting classes etc.

tcbegley:
That's really nice!

I agree that since the interface is in Python it makes sense to stick to snake case as much as possible, though for attributes such as `className` that are present in the existing Dash component libraries we should probably stick to camel case for consistency.

pbugnion:
That makes sense. To summarise:
- we use whatever case Dash uses for attributes with the same name as in Dash.
- we use snake case for other attributes.

PR #17, opened 2018-10-03
Dropdownitem links

tcbegley:
This PR adds better support for links in the `DropdownItem` component.

Currently it looks at the `href` attribute. If this is not present it creates a html button with the correct styling and can be used in callbacks by watching `n_clicks`. If `href` is present, then a `Link` component from `private` is used.

There is one problem with the current implementation which is that when `href` is specified and `Link` gets used, the `onClick` attribute of `DropdownItem` is overwritten. This works well for allowing `DropdownItem` to either redirect to an external page or update the url depending on the form of `href` or the value of `external_link`. However, it does mean that the `n_clicks` and `n_clicks_timestamp` attributes are not properly updated, as `Link` updates these as internal props and the information doesn't flow back up.

There are two possibilities here as far as I can tell.

- Modify `Link` to allow the function that gets called on click to be modified by whichever component creates the `Link`. I guess this would result in complicating `Link`, but since it's an internal component maybe it's ok.
- Decide that there isn't really a reason to want to update the url and also use `n_clicks` to trigger a callback and simply have `n_clicks` not update when `href` has been specified. This does of course feel like a bit of a fudge.

Very happy to take some input on this.

### DIFF
--- a/examples/navitem_test.py
+++ b/examples/navitem_test.py
@@ -1,0 +1,52 @@
+import dash
+import dash_bootstrap_components as dbc
+import dash_html_components as html
+from dash.dependencies import Input, Output
+
+app = dash.Dash()
+
+navbar = dbc.Navbar(
+    brand="Dash Bootstrap components",
+    brand_href="https://github.com/ASIDataScience/dash-bootstrap-components",
+    sticky="top",
+    children=[
+        dbc.NavItem(dbc.NavLink("ASI", href="https://www.asidatascience.com")),
+        dbc.Dropdown(
+            nav=True,
+            inNavbar=True,
+            label="Menu",
+            children=[
+                dbc.DropdownItem("Entry 1", href="https://google.com"),
+                dbc.DropdownItem("Entry 2", href="/test", id="dd-internal"),
+                dbc.DropdownItem(divider=True),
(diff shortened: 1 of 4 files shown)


examples/navitem_test.py
```diff
@@ -0,0 +1,52 @@
...
+)
+
+
+@app.callback(Output("counter", "children"), [Input("dd-button", "n_clicks")])
+def count(n):
+    print(n)
+    return str(n)
+
+
+@app.callback(
```
tcbegley: this callback never gets triggered because `n_clicks` is not updated properly

pbugnion:
This looks great!

`Link` could take a `preRedirect` (or `preOnClick`) function as a prop, and we call this (if not null) as the first item of `onClick` (unless it's disabled)?

Somewhat relatedly, I don't think there's any value in `Link` having `n_clicks` and `n_click_timestamp` props, since the user can't actually instantiate a `Link`. What do you think?

pbugnion:
Sorry -- I definitely didn't mean to close this.

pbugnion:
Should there be a change in the propTypes of `Link`?

tcbegley:
Good point, I think it's worth adding `preOnClick`.

pbugnion:
I think this looks great!

PR #23, opened 2018-10-20
Adding (text) color attribute to components

tcbegley:
This small PR adds text color to the python interface for `Label`, `CardTitle`, `CardSubtitle`, `CardText` and `ListGroupItemText`.

### DIFF
--- a/src/components/Label.js
+++ b/src/components/Label.js
@@ -18,6 +18,7 @@ const Label = props => {
     align,
     xs,
     className,
+    color,
     ...otherProps
   } = props;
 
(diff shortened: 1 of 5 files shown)


src/components/Label.js
```diff
@@ -18,14 +18,19 @@ const Label = props => {
...
 
   // check if column width has been specified, use alignment attribute if so
   const cols = colWidths.filter(colWidth => props[colWidth] || props[colWidth] === '');
 
   const alignClass = align && alignMap[align];
-  const classes = classNames(className, cols.length && alignClass);
+  const classes = classNames(
+    className,
+    cols.length && alignClass,
+    color && `text-${color}`
```
pbugnion: Is it more common to put `className` last? It seems likely that, if the user specifies `className` explicitly, that should just override any other option.

```suggestion
    const classes = classNames(
      cols.length && alignClass,
      color && `text-${color}`,
      className
    )
```

pbugnion: Actually, ignore this -- it's irrelevant since classes are always additive.

pbugnion:
Nice! This looks great!

PR #26, opened 2018-10-24
Add Python linting and reformatting

pbugnion:
This PR adds [flake8](http://flake8.pycqa.org/en/latest/) and [black](https://black.readthedocs.io/en/stable/) to the build pipeline. Black enforces a consistent code style, and flake8 catches common errors like unused imports or unused variables.

I have used the same black configuration as [sherlockml](https://github.com/asidatascience/sherlockml) for consistency across ASI.

The only complexity is that flake8 needs to ignore specific components files in the documentation: the ones that require `app` to be injected when the code is executed. For now, this is just `collapse`.

This also adds a very basic check for syntax errors in the source code for all the versions of Python which I presume we want to target. Since Dash supports Python 2, the cost to supporting it seems minimal.


.travis.yml
```diff
@@ -1,10 +1,25 @@
...
 install:
-  - npm install
+  - "if [[ \"$GROUP\" == js ]] ; then npm -v ; fi"
+  - "if [[ \"$GROUP\" == js ]] ; then npm install ; fi"
+  - "if [[ \"$GROUP\" == python-linting ]] ; then pip install black flake8 ; fi"
 script:
-  - npm run test
+  - "if [[ \"$GROUP\" == js ]] ; then npm run test ; fi"
+  - "if [[ \"$GROUP\" == python-linting ]] ; then black . --check ; fi"
+  - "if [[ \"$GROUP\" == python-linting ]] ; then flake8 ; fi"
```
tcbegley: Here I think we want `flake8 --ignore E203, W503`.  These are, respectively, "whitespace before ':'" and "line break before binary operator" warnings, which `flake8` will complain about, but which aren't PEP8 compliant and `black` does not adhere to. There's a bit of discussion about it in the `black` [README](https://github.com/ambv/black).

.travis.yml
```diff
@@ -1,10 +1,25 @@
...
     - "node_modules"
 install:
-  - npm install
+  - "if [[ \"$GROUP\" == js ]] ; then npm -v ; fi"
+  - "if [[ \"$GROUP\" == js ]] ; then npm install ; fi"
+  - "if [[ \"$GROUP\" == python-linting ]] ; then pip install black flake8 ; fi"
 script:
-  - npm run test
+  - "if [[ \"$GROUP\" == js ]] ; then npm run test ; fi"
+  - "if [[ \"$GROUP\" == python-linting ]] ; then black . --check ; fi"
```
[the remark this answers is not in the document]
pbugnion: We've included the line length in `pyproject.toml`, the configuration file that black looks at, so we don't also need to include it in the command line, afaik.

.travis.yml
pbugnion: Good call. I'll put these in `setup.cfg` (it's better to have these in configuration, rather than just in CI, so that tools that integrate with editors can use them as well.

.travis.yml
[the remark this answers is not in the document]
tcbegley: Ah whoops, I missed that, that's perfect.

.travis.yml
tcbegley: Sounds great!

pbugnion:
I've fixed the flake8 incompatibilities.

setup.cfg
```diff
@@ -0,0 +1,3 @@
+[flake8]
+exclude = node_modules, docs/components/collapse.py
```
tcbegley: Why are we excluding `collapse.py`?

pbugnion: Basically, collapse is not a valid Python file -- it depends on an undefined variable app.

When we build the documentation, we load the code with `exec` and an overriden set of local variables in which we inject `app`.

The complexity around the component pages is that they:
- need to work and actually render the right information
- be somewhat intuitive for the user to understand when they read the code.

pbugnion: https://dash-bootstrap-components.opensource.asidatascience.com/l/components/collapse

tcbegley: That makes sense, thanks for the clarification!

PR #29, opened 2018-10-26
Document layout

tcbegley:
This PR adds the main layout components (`Col` and `Row`) to the documentation.

### DIFF
--- a/docs/assets/docs.css
+++ b/docs/assets/docs.css
@@ -1,91 +1,90 @@
-
 .main-header {
-    background-color: white;
+  background-color: white;
 }
 
 .main-header h1 {
-    margin-bottom: 1rem;
+  margin-bottom: 1rem;
 }
 
 .main-header p.lead {
-    margin-bottom: 1.5rem;
+  margin-bottom: 1.5rem;
 }
 
 .main-header .btn {
-    margin-right: 1rem;
+  margin-right: 1rem;
 }
 
-.main-header .btn + .btn {
-    margin-right: 0;
+.main-header .btn+.btn {
+  margin-right: 0;
 }
 
 .section-header {
-    font-weight: 300;
+  font-weight: 300;
 }
 
 code.hljs {
-    padding: 1em;
-    border-radius: 3px;
+  padding: 1em;
+  border-radius: 3px;
 }
 
 span.hljs-meta {
-    color: #66d9ef;
+  color: #66d9ef;
 }
 
 .demo-layout-container {
-    height: 600px;
-    margin-right: 0rem;
-    margin-left: 0rem;
-    margin-bottom: 3rem;
-    margin-top: 2rem;
-    display: block;
-    box-shadow: 0 0 60px rgba(0, 0, 0, 0.3);
-    border-radius: 6px;
-    overflow: none;
+  height: 600px;
+  margin-right: 0rem;
+  margin-left: 0rem;
+  margin-bottom: 3rem;
+  margin-top: 2rem;
+  display: block;
+  box-shadow: 0 0 60px rgba(0, 0, 0, 0.3);
+  border-radius: 6px;
+  overflow: none;
 }
 
 .demo-layout {
-    width: 100%;
-    height: 100%;
-    display: block;
-    border: 0;
-    border-radius: 6px;
+  width: 100%;
+  height: 100%;
+  display: block;
+  border: 0;
+  border-radius: 6px;
 }
 
 .docs-content {
-    margin-top: 2rem;
+  margin-top: 2rem;
 }
 
 .docs-sidebar .nav-item {
-    padding: 0 .2rem 0 .2rem;
+  padding: 0 .2rem 0 .2rem;
 }
 
 .docs-sidebar .nav-link {
-    color: #888;
-    padding: .3rem .5rem .3rem 1rem;
-    border-left: 2px solid transparent;
+  color: #888;
+  padding: .3rem .5rem .3rem 1rem;
+  border-left: 2px solid transparent;
 }
 
 .docs-sidebar .nav-link:hover {
-    color: #777;
-    border-left: 2px solid #aaa;
+  color: #777;
+  border-left: 2px solid #aaa;
 }
 
 .docs-sidebar .nav-link.active {
-    color: #666;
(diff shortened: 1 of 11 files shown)


pbugnion:
This is great! I think the tests are failing because this needs to be rebased off master? I reformatted some code after you branched off (sorry -- that might result in a somewhat painful merge).

docs/assets/docs.css
```diff
@@ -87,3 +87,35 @@ span.hljs-meta {
 .example-source-container {
     margin-bottom: 2rem;
 }
+
+.layout-demo .row {
```
pbugnion: I have used 4 space layout so far. I actually think 2 is more consistent with what we have, so I suggest changing everything to 2?

tcbegley: Sounds good, I'll add that to this PR

tcbegley:
I think I managed to resolve all the conflicts. The documentation app is running fine for me on this branch. 

@pbugnion if you could try running this and see if it matches your expectations compared to what you most recently merged into master. As far as I can tell they are the same.

Currently the automated API documentation isn't producing particularly helpful results for width arguments in `Col`. I think likely because those attributes have a relatively specialised PropType. I'll take a look at that to see if there's an easy fix before merging this.

pbugnion:
I think this looks great! I could get the documentation deployed etc.

I've had a go at doing the rebase myself and came up with a somewhat cleaner history (see PR #32).

I can see the following minor issues:
- At the moment, the API documentation doesn't actually say what component is being described. While that was fairly obvious for single component pages, it's not obvious for `Row` and `Column`. Maybe we can make the title of the `ApiDoc` component customisable?
- Should we change `noGutters` to `no_gutters`? Or was there a reason this was kept as camel case?
- I think Dash should be able to resolve the prop type for `md`, `xs` etc, based on https://github.com/plotly/dash/blob/master/dash/development/base_component.py#L727. Not sure why it isn't working. Anyway, maybe we could add more documentation in the free text field, and some examples? It'd be fairly repetitive, but probably better than not having documentation?

tcbegley:
-  A customisable title for `ApiDoc` is a nice idea. I could have a go at adding that on this PR?
- I have on my todo list to go through and change all camel case attributes (that don't already exist in the core Dash libraries) to snake case. I was going to wait until the Javascript reformatting has been merged.
- I think some more documentation would definitely be good. I was thinking it would also be good if we came up with a nice way of adding some text between the examples, so that for example on the layout page we can explain some key ideas with words in between the examples?

pbugnion:
> A customisable title for ApiDoc is a nice idea. I could have a go at adding that on this PR?

Yes, that would be great!

> I have on my todo list to go through and change all camel case attributes (that don't already exist in the core Dash libraries) to snake case. I was going to wait until the Javascript reformatting has been merged.

Agreed. Let's merge this first, then.

> I think some more documentation would definitely be good. I was thinking it would also be good if we came up with a nice way of adding some text between the examples, so that for example on the layout page we can explain some key ideas with words in between the examples?

Yes, that sounds good. Do you think we need anything other than just putting divs in between the HighlightedSource / ExampleContainer objects?

docs/components_page/page.py
```diff
@@ -94,6 +118,36 @@ def __init__(self, app):
                 HighlightedSource(collapse_source),
                 ApiDoc(component_metadata.get("src/components/Collapse.js")),
             ],
+            "layout": html.Div(
+                [
+                    layout_simple,
```
pbugnion: I think this (and all the other `layout_`) should be wrapped in the `ExampleContainer` component. It imposes some margins etc.

tcbegley:
> Yes, that sounds good. Do you think we need anything other than just putting divs in between the HighlightedSource / ExampleContainer objects?

I think that's fine, though I wonder if the text itself should be read in from external files somehow rather than being embedded in the source code? Depending on how long the text is the source could get pretty bloated. What do you think?

docs/components_page/page.py
tcbegley: Good point, I've added those.

pbugnion:
> Depending on how long the text is the source could get pretty bloated.

I think `page.py` might get pretty bloated. We could consider moving everything related to a given component into its separate source file, and just exporting a list from that file. For instance, you could have a `layout.py` (in the same directory as `page.py`) with:

```py
# layout.py

layout_simple_source = (COMPONENTS / "layout" / "simple.py").open().read()
layout_width_source = (COMPONENTS / "layout" / "width.py").open().read()
layout_order_offset_source = (
    (COMPONENTS / "layout" / "order_offset.py").open().read()
)
layout_breakpoints_source = (
    (COMPONENTS / "layout" / "breakpoints.py").open().read()
)
layout_no_gutters_source = (
    (COMPONENTS / "layout" / "no_gutters.py").open().read()
)
layout_vertical_source = (COMPONENTS / "layout" / "vertical.py").open().read()
layout_horizontal_source = (
    (COMPONENTS / "layout" / "horizontal.py").open().read()
)

component = html.Div(
                [
                    layout_simple,
                    html.Div("some custom text"),
                    HighlightedSource(layout_simple_source),
                    layout_width,
                    HighlightedSource(layout_width_source),
                    layout_order_offset,
                    HighlightedSource(layout_order_offset_source),
                    layout_breakpoints,
                    HighlightedSource(layout_breakpoints_source),
                    layout_no_gutters,
                    HighlightedSource(layout_no_gutters_source),
                    html.Div(
                        [
                            layout_vertical,
                            HighlightedSource(layout_vertical_source),
                        ],
                        className="pad-row",
                    ),
                    layout_horizontal,
                    HighlightedSource(layout_horizontal_source),
                    ApiDoc(
                        component_metadata.get("src/components/layout/Row.js")
                    ),
                    ApiDoc(
                        component_metadata.get("src/components/layout/Col.js")
                    ),
                ],
                className="layout-demo",
)
```

Then, in `page.py`, we just need to:

```py
from .layout import component as layout_component

# ...

self._component_bodies = {
    "layout": layout_component
}
```

tcbegley:
Nice, I think that’s a good approach.

Shall we leave that for another PR and get this merged?

pbugnion:
Merging this and adding additional documentation later sounds like a good plan. In this case, this looks ready to merge, unless there's something else you think we should add?

PR #31, opened 2018-10-27
Javascript reformatting

pbugnion:
This PR adds a configuration for prettier. I have chosen the [configuration](https://github.com/ASIDataScience/dash-bootstrap-components/blob/f0b7be117bfddce4b44f745b9fe1e94a22de9513/.prettierrc) to lead to something somewhat similar to our current format. Other options would be to use the same configuration as [dash-core-components](https://github.com/plotly/dash-core-components/blob/5b31e5f7f6ee00fabffc0b78bfedf620f74aeb8f/.prettierrc).

To reformat the code, `npm run format`.

This PR also adds a check that the entire code is formatted correctly to the CI pipeline.


tcbegley:
This looks great! Thanks for setting it up. Happy for you to merge when ready.

PR #32, opened 2018-10-30
Document layout rebased

pbugnion:
This PR is equivalent to PR #29 , but I have rebased it to avoid re-committing commits e00bddb, 595227d and 73f37de9. The commits to fix the rebase (f04bb7f1be3 and a7752bb80) then naturally drop out. This may come across as fairly pedantic, but I think it makes it somewhat easier to maintain.

Apart from resulting in nicer looking commit graphs, a cleaner history makes it:
- easier to run git bisect to understand where regressions were introduced
- easier to understand the scope of changes in a given PR.

Tom -- those commits are now listed with you as the author and me as the committer -- while GitHub recognizes the difference, most programs that show history just show the author.

Deployed [here](https://dash-bootstrap-components-documentation.app.asi.sherlockml.net/l/components/layout).

If you're happy that this has everything, I suggest that I force push this branch back onto the original PR.

### DIFF
--- a/docs/assets/docs.css
+++ b/docs/assets/docs.css
@@ -1,91 +1,90 @@
-
 .main-header {
-    background-color: white;
+  background-color: white;
 }
 
 .main-header h1 {
-    margin-bottom: 1rem;
+  margin-bottom: 1rem;
 }
 
 .main-header p.lead {
-    margin-bottom: 1.5rem;
+  margin-bottom: 1.5rem;
 }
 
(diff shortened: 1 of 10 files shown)


tcbegley:
Thanks for this @pbugnion! I certainly managed to get myself into a bit of a mess originally.

This looks good to me, I think we can push it onto the original PR.

pbugnion:
Done, thanks.

PR #33, opened 2018-10-31
Use snake case

tcbegley:
This PR changes all camel case attributes (that don't already exist in the core Dash libraries) to snake case. I have also updated the examples, demo and docs.


pbugnion:
This looks great! Merge when you're ready!

PR #35, opened 2018-11-02
Add documentation for cards

pbugnion:
This PR adds documentation for cards. It is deployed on [SherlockML](https://dash-bootstrap-components-documentation.app.asi.sherlockml.net/l/components/cards).

`page.py` is becoming a bit unwieldy, so I might, as a separate PR, try to split it out into a page for each component.

PR #36, opened 2018-11-03
Refactor creating the layout for the content of each component

pbugnion:
As discussed in PRs #35 and #29 , the current structure where we define the entire components page in a single source file, `components_page/page.py`, gets unwieldy. This PR refactors the structure of that page so it just loads a page for each component, `components_page/<component>_content.py`.

This leads to a much more extensible structure overall. It makes it more natural to add additional documentation sections for individual components.

There is no visible change to the documentation. I have verified that the documentation still works as before.


tcbegley:
This is great, thanks for setting it up!

PR #37, opened 2018-11-03
Document fade

pbugnion:
This adds documentation for `Fade`.

This should be reviewed after PR #36 .

### DIFF
--- a/docs/components_page/components/fade.py
+++ b/docs/components_page/components/fade.py
@@ -1,0 +1,29 @@
+import dash_bootstrap_components as dbc
+import dash_html_components as html
+from dash.dependencies import Input, Output, State
+
+fade = html.Div(
+    [
+        html.H2("Fade"),
+        dbc.Button(
+            "Toggle fade", id="fade-button", style={"margin-bottom": "1rem"}
+        ),
+        dbc.Fade(
+            dbc.Card(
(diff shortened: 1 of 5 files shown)


tcbegley:
This is looking good!

- Looks like we need to add `fade.py` to the list of files for flake8 to ignore in `setup.cfg`, I suspect you probably already knew that.
- I noticed that in the example, the first time you click the toggle button nothing happens, which is because `in`, not being initialised, is initially `None` and so gets toggled to `True` on its first click. However this then lead me to realise the bigger problem that you can't specify an initial value for `in` without python throwing a syntax error. All of which is to say that I think we should change the prop `in` to `is_in`.

tcbegley:
I have addressed the second bullet in PR #38. I branched off master so that PR does not include any changes to the Fade documentation. Best thing would probably be to merge that before this if we're happy with it and make the corresponding changes here?

pbugnion:
Well spotted, thanks a lot for fixing this.

pbugnion:
This is ready for review, I think. The app is deployed on [SherlockML](https://dash-bootstrap-components-documentation.app.asi.sherlockml.net/l/components/fade). Since the app just installs 0.1.2, it doesn't include the API documentation changes that I've made to `Fade.js`.

tcbegley:
Looks good to me!

PR #39, opened 2018-11-04
Rename Dropdown to DropdownMenu

tcbegley:
This PR renames `Dropdown` to `DropdownMenu` and `DropdownItem` to `DropdownMenuItem`.

This is to avoid a clash with the `Dropdown` component in `dash-core-components` which we may in future want to wrap and add additional styling to.


pbugnion:
Great, thank you!

PR #38, opened 2018-11-04
Rename Fade in prop

tcbegley:
This PR renames the `in` prop of the `Fade` component to `is_in`. This allows the initial value to be specified in the python interface (previously a syntax error was thrown because `in` is of course a reserved word in Python), and also more closely shadows the `is_open` prop of `Collapse`.


pbugnion:
This is great, thanks!

PR #40, opened 2018-11-05
Document Tooltip and Popover

tcbegley:
This PR adds documentation for the `Popover` and `Tooltip` components.


pbugnion:
Nice, this looks great!

Issue #21, opened 2018-10-17
Bootstrap styling for DatePicker components


tcbegley:
`DatePickerSingle` and `DatePickerRange` work well inside `FormGroup` components, but the default styling makes them look a bit out of place when next to Bootstrap themed inputs. For example, they are taller than Bootstrap inputs and don't have rounded corners. It would be good to look into whether we can bundle some CSS that unifies everything.

tcbegley:
This is being addressed by #43

PR #43, opened 2018-11-06
Add bootstrap compatible date pickers

tcbegley:
This PR adds `DatePickerSingle` `DatePickerRange` components which which shadow the implementations in Dash core components, but with the styles modified to fit with bootstrap components.

In addition to this it adds a `bs_size` prop which allows you to control the size of the components in the same way that bootstrap allows you to size form controls, see [here](https://getbootstrap.com/docs/4.1/components/forms/#sizing)

This closes issue #21. You can see it working in `examples/date_picker.py`.

Here's a screenshot of the different sizes in action

image

### DIFF
--- a/dash_bootstrap_components/__init__.py
+++ b/dash_bootstrap_components/__init__.py
@@ -18,12 +18,7 @@ _js_dist = [
     }
 ]
 
-_css_dist = [
-    {
-        "relative_package_path": "react-dates-shim.css",
-        "namespace": "dash_bootstrap_components",
-    }
-]
+_css_dist = []
 
 
 def _setup_js_components(module, path_to_metadata):
--- a/dash_bootstrap_components/react-dates-shim.css
+++ b/dash_bootstrap_components/react-dates-shim.css
@@ -1,3 +1,0 @@
-.form-group > .SingleDatePicker, .form-group > .DateRangePicker {
-  display: table;
-}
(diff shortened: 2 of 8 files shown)


tcbegley:
I have added a `disabled_days` prop that allows you to optionally pass a list of days to disable in addition to those that fall outside the range specified by `min_date_allowed` and `max_date_allowed`

SundareshPrasanna:
Hey daterangepicker in dcc is still difficult to modify (in terms of styling) any way to bring a more friendly version to dbc?

tcbegley:
Hey @SundareshPrasanna 

We've been working on some stylesheets that add Bootstrap styling to the date pickers in *dash-core-components*. The plan is to host them on our own CDN and link through the `dbc.themes` module, but for now you can obtain copies I compiled from [here](https://github.com/tcbegley/dash-bootstrap-css/tree/master/dist)

Download the theme you want and link to your app, for example by placing it in your apps `assets/` folder, then make sure that any date picker you want to add the bootstrap styles to is the child of an element with `className="dash-bootstrap"` (doesn't have to be an immediate parent). For example

```python
import dash

app = dash.Dash()

app.layout = html.Div([dcc.DatePickerSingle()], className="dash-bootstrap")

if __name__ == "__main__":
    app.run_server(debug=True)
```

Running that code with the darkly css in my `assets/` folder I get something like this:

image

Let me know how you get on if you try it, and especially if you encounter bugs!

PR #45, opened 2018-11-07
ApiDoc updates

tcbegley:
This PR adds the component name to available events in the auto-generated api docs.

I also plan to investigate why the prop types are not being displayed properly for `Col` sizing props.

### DIFF
--- a/docs/components_page/api_doc.py
+++ b/docs/components_page/api_doc.py
@@ -13,7 +13,7 @@ def ApiDoc(component_metadata, component_name=None):
     component_props = component_metadata.get("props", {})
     return html.Div(
         ArgumentsList(component_props, component_name)
-        + EventsList(component_props),
+        + EventsList(component_props, component_name),
         className="api-documentation",
     )
 
@@ -49,7 +49,11 @@ def Argument(argument_name, argument_metadata):
     )
 
 
-def EventsList(component_props):
+def EventsList(component_props, component_name):
+    if component_name is not None:
+        heading = f"Available events for {component_name}"
+    else:
+        heading = "Available events"
     events = parse_events(component_props)
     if not events:
         return []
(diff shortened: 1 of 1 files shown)


pbugnion:
This looks great so far.

tcbegley:
@pbugnion I looked into why the prop types for `Col` are not displaying properly.

[In the source for Col](https://github.com/ASIDataScience/dash-bootstrap-components/blob/master/src/components/layout/Col.js#L24-L45) we define a custom prop type `columnProps` as is done in the Reactstrap source. Mainly because each of the props `width`, `xs`, `sm`, `md`, `lg`, and `xl` all take the same relatively complicated union of different prop types, which would result in a decent amount of bloat if written out explicitly 6 times.

Then [here](https://github.com/ASIDataScience/dash-bootstrap-components/blob/master/docs/components_page/api_doc.py#L41-L42) the type gets extracted as `custom` which the `dash` function `js_to_py_type` doesn't know what to do with.

It's pretty easy to insert a check at that point, and return a custom description of the columnProps type, however this only serves to fix the `ApiDoc` function. We're still left with the problem that the docstrings of the underlying python object don't have a good description of the expected prop type.

Is is possible to configure the way that the React docstrings get generated such that the generated metadata is equivalent to what we would get if we went through and replaced `columnProps` with an explicit description of the prop types? Or alternatively is adding that extra bloat worth it to get nice auto-generated docstrings and documentation?

Very interested to hear your thoughts.

pbugnion:
I think the underlying problem is in react-docgen. Instead of resolving the variable `columnProps`, it just bails and assumes it's a custom definition.

Having spent a bit of time looking around react-docgen, we may be able to resolve this by either using another handler that extracts `columnProps` correctly, or defining our own handler. Obviously this is a non-trivial amount of work.

I think we should merge this PR as is, and introduce a new issue for documenting column props correctly. In the meantime, we can just document the type manually in the prop description.

tcbegley:
Great, I'll merge this, and have created #67 and #68 to track docstring improvement plans.

PR #48, opened 2018-11-10
Fix outdated demo and verify demo builds in CI

pbugnion:
The JS demo was broken because it still used `<Dropdown />` instead of `<DropdownMenu />`. This fixes the demo and adds a basic check that the demo builds to the CI pipeline to avoid regressions.


tcbegley:
Ah good catch, I must have missed that, apologies. Good idea to add the demo build to the CI pipeline

PR #49, opened 2018-11-10
Fix hot reloading when demo runs behind a proxy

pbugnion:
Prior to this PR, `npm run demo` created an app where hot-reloading did not work if the app was behind a proxy, because the browser client looked for the backend at localhost, rather than using window.location. To work at all, we disabled the host check, which, in principle, opened the backend to phishing attacks [1].

Now, we allow configuring the host header for the frontend using the DEMO_PUBLIC_HOST environment variable [2]. If that variable is set, it will be used by the frontend and by the host check in the backend. If not, we fall back to the webpack default (localhost:<port>, in most occasions).

[1] These would not work on SherlockML because we have an additional layer of security that verifies the host header.

[2] To make this work on SherlockML, write the following in your shell:

```
export SHERLOCKML_PUBLIC_HOST=cube-"$SHERLOCKML_SERVER_ID".workspace.asi.sherlockml.net
```

Builds on PR #48.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "build-dist": "npm run clean-lib && npm run extract-metadata && webpack --config=webpack.config.dist.js",
     "clean-lib": "mkdirp lib && rimraf lib/*",
     "copy-lib": "copyfiles -u 1 lib/* dash_bootstrap_components",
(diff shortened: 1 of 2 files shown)


tcbegley:
This is great and will make development so much nicer, thanks for adding it!

PR #51, opened 2018-11-11
Document input group

pbugnion:
This adds documentation for input groups. I'd originally planned to document `dbc.Input` as part of this, but realised that that was probably sufficiently large to warrant its own separate documentation.

This is rebased on top of #59.

I can't deploy this very easily since it requires building the bundle, but this is what the documentation looks like:

dbc-input-group-screenshot


tcbegley:
This looks great! Thanks @pbugnion 

I think you're right that `Input` deserves its own page.

PR #55, opened 2018-11-13
Document dropdown menu

pbugnion:
This PR documents the DropdownMenu component. It builds off PR #64  .

This is what the documentation looks like:

dbc-dropdown-menu

### DIFF
--- a/src/components/DropdownMenu.js
+++ b/src/components/DropdownMenu.js
@@ -121,8 +121,8 @@ DropdownMenu.propTypes = {
   caret: PropTypes.bool,
 
   /**
-   * Size of the dropdown. `sm` corresponds to small, `md` to medium
-   * and `lg` to large.
+   * Size of the dropdown. 'sm' corresponds to small, 'md' to medium
+   * and 'lg' to large.
    */
   bs_size: PropTypes.oneOf(['sm', 'md', 'lg'])
 };
--- a/src/components/DropdownMenu.js
+++ b/src/components/DropdownMenu.js
@@ -67,7 +67,7 @@ DropdownMenu.propTypes = {
   id: PropTypes.string,
 
   /**
-   * The children of this component
+   * The children of this component.
    */
   children: PropTypes.node,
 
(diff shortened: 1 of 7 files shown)


tcbegley:
This is already looking great! 

Looks like the `size` prop isn't currently exposed in the Python API for `DropdownMenu`, this would be a good place to add it I guess. Should we use `bs_size` as in `Input` and `DatePicker*` for consistency?

There's also something pretty strange happening with the dropup, seems to be a problem with the underlying component itself.

pbugnion:
> Looks like the size prop isn't currently exposed in the Python API for DropdownMenu, this would be a good place to add it I guess. Should we use bs_size as in Input and DatePicker* for consistency?

Agreed. I've done this in PR #61 , which this is now rebased on top of.

> There's also something pretty strange happening with the dropup, seems to be a problem with the underlying component itself.

Well spotted. I had a brief look at this and it wasn't completely obvious what we were doing wrong (if anything). I've opened issue #62 to track this.

pbugnion:
I think this is now ready for review.

src/components/DropdownMenu.js
```diff
@@ -80,17 +82,17 @@ DropdownMenu.propTypes = {
...
-   * Label for the toggle
+   * Label for the toggle.
    */
   label: PropTypes.string,
 
   /**
-   * Direction in which to expand the dropdown.
+   * Direction in which to expand the dropdown. Default: 'down'.
    */
   direction: PropTypes.oneOf(['up', 'down', 'left', 'right']),
```
tcbegley: Shall we change this to just 'down', 'left' or 'right' as discussed in #62 as part of this PR?

tcbegley:
This looks great! The only question I have is whether we should implement the temporary fix for #62 as part of this PR as commented inline.

pbugnion:
> The only question I have is whether we should implement the temporary fix for #62 as part of this PR as commented inline.

That definitely makes sense. I've pushed PR #64 separately and rebased this on top of that.

tcbegley:
Looks great!

PR #57, opened 2018-11-13
Document table

tcbegley:
This PR adds a simple `Table` example to the docs.


pbugnion:
This is great! 

Separately, we could consider a python method to construct a table from a dataframe.

tcbegley:
Good idea, I actually have something like this lying around, but will add it in a separate PR. For now I've logged it in issue #66 and will get this merged.

PR #58, opened 2018-11-13
Document Tabs

tcbegley:
This PR adds documentation for tabs.

@pbugnion I would be particularly interested if you could take a look at the modifications in `helpers.py`, I had to pass some module imports into `exec` in order to get the callbacks to work, previously it was complaining about `dbc` not being recognised etc. What I've got works, but there may be better ways to do it.


docs/components_page/components/tabs/simple.py
```diff
@@ -0,0 +1,35 @@
...
+)
+
+tab2_content = dbc.Card(
+    dbc.CardBody(
+        [
+            dbc.CardText("This is tab 2!"),
+            dbc.Button("Don't click here", color="danger"),
+        ]
+    ),
+    # add small margin to the top of the card
```
pbugnion: I'm not sure this comment (and the above) add that much value. At least, we should be able to have just one.

docs/components_page/tabs_content.py
```diff
@@ -0,0 +1,55 @@
...
+        ExampleContainer(tabs_simple),
+        HighlightedSource(tabs_simple_source),
+        active_tab_blurb,
+        ExampleContainer(
+            load_source_with_app(
+                app,
+                tabs_active_source,
+                "tabs",
+                {
+                    "dbc": dbc,
```
pbugnion: When I replicated this, I didn't need to explicitly inject `dbc`.

docs/components_page/helpers.py
```diff
@@ -15,13 +15,18 @@ def ExampleContainer(component):
     return html.Div(component, className="example-container")
 
 
-def load_source_with_app(app, source, component_name):
+def load_source_with_app(
+    app, source, component_name, globals_dict=None, locals_dict=None
```
pbugnion: Are we ever using `locals_dict`?

pbugnion: In fact, would it be better to inject additional local variables in `locals_dict`?

We could consider changing `load_source_with_app` to `load_source_with_locals` and folding passing `app` into the local variables.

This would then be called with:

```py
load_source_with_locals(
    tabs_active_source, 
    {'app': app, 'tab1_content': tab1_content, 'tab2_content': tab2_content}
)

pbugnion:
I think this looks great! I've made a couple of comments inline, specially around helpers.

docs/components_page/tabs_content.py
tcbegley: Strange, I was definitely getting a `NameError` when I tried. I'll take another look!

docs/components_page/helpers.py
tcbegley: I like this, I'll have a go at making the relevant changes

tcbegley:
Thanks for the review @pbugnion! I agree that we should be careful to keep snippets relatively self-contained. I'll take another look at the `load_source_with_app` function and implement your suggestions.

tcbegley:
@pbugnion this is now ready for another review

The main outstanding question I have is on naming. The function name `load_source_with_locals` doesn't quite sit right with its argument `globals_dict` which takes its name from `exec`'s `globals` parameter. Any preference on how best to resolve? Or whether you think there is anything to resolve?

pbugnion:
I was originally thinking of passing the variables as locals to exec, not globals, hence the name. According to the docstring on exec, though:

> The globals must be a dictionary and locals can be any mapping,
defaulting to the current globals and locals.
> If only globals is given, locals defaults to it.

Thus, we can't pass just locals. I think we should just rename the function to `load_source_with_globals`, or `load_source_with_environment`?

tcbegley:
I actually tried something very similar to this originally, and checked the above snippet just to be sure, but if you do that the documentation page for tabs throws a name error saying `tab1_content` is not defined. For a reason I haven't been able to figure out yet that needs to be passed to `globals` in `exec` to work. In which case would the best thing be to pass everything in to globals, leave the current implementation more or less as is but rename to `load_source_with_environment` as suggested in a previous comment?

pbugnion:
Ah that makes sense. Not sure why it's not working, but my understanding of `exec` is fairly limited.

> In which case would the best thing be to pass everything in to globals, leave the current implementation more or less as is but rename to load_source_with_environment as suggested in a previous comment?

I think that make sense, yes.

Thanks for tackling this!

pbugnion:
This looks perfect, thank you!

Issue #62, opened 2018-11-17
Transformation is not set correctly for dropdown menu when direction is `up`


pbugnion:
When creating a dropdown menu with `direction == 'up'`, items do not show correctly. In particular, the style `bottom` is not set to `auto` and the transformation that is supposed to push the items above the dropdown is wrong.

<img width="303" alt="screen shot 2018-11-17 at 14 53 55" src="https://user-images.githubusercontent.com/1392879/48662342-c8c30800-ea78-11e8-942a-cd17224e4ad2.png">

pbugnion:
## Cause

tl;dr this is caused by dash using react 15, not 16.

Long version:

Setting `direction = "up"` as a prop works in `Demo.js`. 

By looking at the [reactstrap source code](https://github.com/reactstrap/reactstrap/blob/214da8c6fb02ccb1198444a378e85851f366b87a/src/DropdownMenu.js#L71), reactstrap injects a prop called `x-placement` in `RSDropdownMenu` that changes depending on the dropdown direction.

[Bootstrap](https://github.com/twbs/bootstrap/blob/d0b7eb1111c7b3c3b6f1291b5e553404ecac1c3a/scss/_dropdown.scss#L109) then uses a selector based on `x-placement` to move the popup to the correct place.

The `x-placement` attribute is not a standard DOM attribute. It's just a bootstrap thing. In HTML, I believe you can inject arbitrary attributes on divs, but the expectation is that you use [data attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes). Your custom attributes should start with `data-`. Bootstrap ignores this convention, presumably because it predates it.

Up to React 15, any extra attributes passed to `<div />` (or any other JSX DOM elements) just gets silently stripped out. In React 16, they get passed through. [See here for changelog](https://reactjs.org/blog/2017/09/08/dom-attributes-in-react-16.html). 

Thus, in `Demo.js` which uses React 16, the `x-placement` attribute makes its way into the DOM, where it is picked up by the bootstrap CSS.

<img width="924" alt="screenshot 2018-11-18 at 10 57 14" src="https://user-images.githubusercontent.com/1392879/48671475-bbf9ef00-eb20-11e8-98bb-994548949ce0.png">

[...]

PR #64, opened 2018-11-19
Disable opening dropdown menus upwards

pbugnion:
Following the discussion in issue #62 , this disables setting the dropdown direction to `up`.

Since dash does not explicitly validate the `oneOf` prop type, this is not actually enforced at the moment: there is nothing stopping a user from passing `up` to the `direction` prop. The only value of this PR is for the docstrings.

If we want to enforce disallowing up, we would have to catch it in `render` and display a console warning or something of the sort. Obviously, in an ideal world, Dash would enforce the `oneOf` prop type on the Python side.

### DIFF
--- a/src/components/DropdownMenu.js
+++ b/src/components/DropdownMenu.js
@@ -87,9 +87,10 @@ DropdownMenu.propTypes = {
   label: PropTypes.string,
 
   /**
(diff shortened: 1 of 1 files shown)


tcbegley:
Looks good to me. I think it's unnecessary to enforce anything in this case.

PR #76, opened 2018-11-24
Check that twine and npm are present on the system before starting releasing

pbugnion:
Prior to this PR, if twine was missing, `invoke prerelease` or `invoke release` would spend a few minutes building the JS bundle and then crash. This introduces a minimal check that the release can actually happen.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -1,8 +1,8 @@
 import os
 import tempfile
 from pathlib import Path
-from subprocess import call
 from shutil import which
+from subprocess import call
 
 import semver
 from invoke import run as invoke_run
--- a/tasks.py
+++ b/tasks.py
@@ -2,6 +2,7 @@ import os
 import tempfile
 from pathlib import Path
 from subprocess import call
+from shutil import which
 
 import semver
 from invoke import run as invoke_run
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -190,6 +192,23 @@ def open_editor(initial_message):
     return lines
 
 
+def check_prerequisites():
+    result = invoke_run("which twine", hide=True)
```
srstevenson: `which` isn't very portable: a more Pythonic alternative might be:
```python
from distutils.spawn import find_executable


if find_executable("twine") is None:
    error(
        "twine executable not found. "
        "You must have twine to run this command."
    )
    exit(127)
```

tasks.py
```diff
@@ -190,6 +192,23 @@ def open_editor(initial_message):
...
+    if result.exited != 0:
+        error(
+            "twine executable not found. "
+            "You must have twine to run this command."
+        )
+        exit(result.exited)
+    result = invoke_run("which npm", hide=True)
```
srstevenson: Ditto.

tasks.py
pbugnion: ah nice, thank you!

PR #82, opened 2018-11-24
Support internal and external links in list group items

tcbegley:
This PR adds support for internal and external links to `ListGroupItem`, the implementation is very much the same as that for `DropdownMenuItem`.

The changes can be tested with the following simple app

```python
import dash

app = dash.Dash()

app.layout = dbc.Container(
    [
        dbc.ListGroup(
            [
                dbc.ListGroupItem("Normal item"),
                dbc.ListGroupItem("Internal link", href="/internal"),
                dbc.ListGroupItem(
                    "Relative link", href="/relative", external_link=True
                ),
                dbc.ListGroupItem("External link", href="https://google.com"),
                dbc.ListGroupItem("Button", id="button-item", n_clicks=0),
                dbc.ListGroupItem(
                    [
                        dbc.ListGroupItemHeading("Item with heading"),
                        dbc.ListGroupItemText("and text"),
                    ],
                    href="/internal2",
                ),
            ]
        ),
        html.P(id="counter"),
    ]
)


@app.callback(
    Output("counter", "children"), [Input("button-item", "n_clicks")]
)
def count(n):
    return f"Button has been clicked {n} times"


if __name__ == "__main__":
    app.run_server(port=8888)

```

### DIFF
--- a/src/components/listgroup/ListGroupItem.js
+++ b/src/components/listgroup/ListGroupItem.js
@@ -22,10 +22,11 @@ class ListGroupItem extends React.Component {
   }
 
   render() {
-    let {children, href, ...otherProps} = this.props;
+    let {children, ...otherProps} = this.props;
+    const {href, disabled} = this.props;
     otherProps[href ? 'preOnClick' : 'onClick'] = this.incrementClicks;
     return (
-      <RSListGroupItem tag={href ? Link : 'li'} href={href} {...otherProps}>
+      <RSListGroupItem tag={href && !disabled ? Link : 'li'} {...otherProps}>
         {children}
       </RSListGroupItem>
     );
--- a/src/components/listgroup/ListGroupItem.js
+++ b/src/components/listgroup/ListGroupItem.js
@@ -11,22 +11,20 @@ class ListGroupItem extends React.Component {
   }
 
   incrementClicks() {
-    if (!this.props.disabled) {
-      if (this.props.setProps) {
-        this.props.setProps({
-          n_clicks: this.props.n_clicks + 1,
-          n_clicks_timestamp: Date.now()
-        });
-      }
+    if (!this.props.disabled && this.props.setProps) {
+      this.props.setProps({
+        n_clicks: this.props.n_clicks + 1,
+        n_clicks_timestamp: Date.now()
+      });
     }
   }
 
   render() {
     let {children, ...otherProps} = this.props;
-    const {href, disabled} = this.props;
-    otherProps[href ? 'preOnClick' : 'onClick'] = this.incrementClicks;
+    const useLink = this.props.href && !this.props.disabled;
+    otherProps[useLink ? 'preOnClick' : 'onClick'] = this.incrementClicks;
     return (
-      <RSListGroupItem tag={href && !disabled ? Link : 'li'} {...otherProps}>
+      <RSListGroupItem tag={useLink ? Link : 'li'} {...otherProps}>
         {children}
       </RSListGroupItem>
     );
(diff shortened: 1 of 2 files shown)


tcbegley:
Marked this as WIP for now as the `disabled` prop is currently not preventing links from being clickable.

tcbegley:
Resolved the issue with links not being disabled. The problem was that the reactstrap `ListGroupItem` component was eating the disabled prop and preventing it from getting passed to our `Link` component, so I figured that the easiest way to resolve it was to simply create a list item rather than a link if `disabled` is set to true.

You can test disabling and enabling all the different links with this modified example:

```python
import dash

app = dash.Dash()

items = ["normal", "internal", "relative", "external", "button", "internal2"]

app.layout = dbc.Container(
    [
        dbc.ListGroup(
            [
                dbc.ListGroupItem("Normal item", id="normal"),
                dbc.ListGroupItem(
                    "Internal link", href="/internal", id="internal"
                ),
                dbc.ListGroupItem(
                    "Relative link",
                    href="/relative",
                    external_link=True,
                    id="relative",
                ),
                dbc.ListGroupItem(
                    "External link", href="https://google.com", id="external"
                ),
                dbc.ListGroupItem("Button", id="button", n_clicks=0),
                dbc.ListGroupItem(
                    [
                        dbc.ListGroupItemHeading("Item with heading"),
                        dbc.ListGroupItemText("and text"),
                    ],
                    href="/internal2",
                    id="internal2",
                ),
            ]
        ),
        html.P(id="counter"),
        dcc.Dropdown(
            id="dropdown",
            options=[
                {"label": item.capitalize(), "value": item} for item in items
            ],
            multi=True,
            value=[],
        ),
    ]
)


@app.callback(Output("counter", "children"), [Input("button", "n_clicks")])
def count(n):
    return f"Button has been clicked {n} times"


def toggle_builder(name):
    def toggler(names):
        if name in names:
            return True
        return False
    return toggler


for item in items:
    app.callback(Output(item, "disabled"), [Input("dropdown", "value")])(
        toggle_builder(item)
    )


if __name__ == "__main__":
```

src/components/listgroup/ListGroupItem.js
```diff
@@ -1,10 +1,37 @@
...
+class ListGroupItem extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.incrementClicks = this.incrementClicks.bind(this);
+  }
+
+  incrementClicks() {
+    if (!this.props.disabled) {
+      if (this.props.setProps) {
```
pbugnion: I feel like it's more common to group if statements:

```js
if (!this.props.disabled && this.props.setProps) {
  // ...
}
```

pbugnion: I don't have particularly strong opinions though.

src/components/listgroup/ListGroupItem.js
```diff
@@ -1,10 +1,37 @@
...
+      }
+    }
+  }
+
+  render() {
+    let {children, ...otherProps} = this.props;
+    const {href, disabled} = this.props;
+    otherProps[href ? 'preOnClick' : 'onClick'] = this.incrementClicks;
+    return (
+      <RSListGroupItem tag={href && !disabled ? Link : 'li'} {...otherProps}>
```
pbugnion: The associativity between `&&` and `?` are not obvious. Would it be clearer with brackets?

```
(href && !disabled) ? Link : 'li'
```

pbugnion:
This looks great, thanks! I made a couple of comments inline.

src/components/listgroup/ListGroupItem.js
pbugnion: In fact, since in PR #83 you use a temporary variable called `useLink`, is there value in using the same construct here?

Using the same syntax when code flows are similar helps a lot with maintenance.

PR #83, opened 2018-11-24
Fix hover behaviour of disabled DropdownMenuItems

tcbegley:
This PR improves the hover behaviour of disabled `DropdownMenuItem` components, by preventing `<a>` elements being used to render disabled items. This previously resulted in pointer cursor appearing when hovering, suggesting the item was clickable when it isn't.


pbugnion:
Looks great!

PR #84, opened 2018-11-24
Document ListGroup

tcbegley:
This PR adds documentation for ListGroup and associated components. It branches off #82 so won't run with any of the released versions of `dash-bootstrap-components` as it uses the `n_clicks` prop of `ListGroupItem` which didn't previously exist. Here's a screenshot of what it looks like

screencapture-127-0-0-1-8888-l-components-list_group-2018-11-24-19_06_48


pbugnion:
Nice! This looks great, thank you!

PR #85, opened 2018-11-24
Replace Path.open().read() with Path.read_text()

srstevenson:
`Path.read_text()` returns the decoded contents of the pointed-to file as a string, and also handles closing the file after reading which previously wasn't being done.


pbugnion:
Ah nice, thanks!

PR #86, opened 2018-11-26
Docs restructuring

tcbegley:
This pull request makes a few changes to the components page of the docs.

Visible changes:

- Rounded corners on code snippets to mirror the styling on the home page of the docs.
- Added a border to examples container to more clearly indicate what is created by the code snippet
- Moved section headings out of the code snippets. It seems to me that the code snippets make sense as minimal copyable examples, and the section headings are probably not something people would want to copy. Moving the headings out also has the benefit of reducing the complexity of the code snippets (in a number of cases I was able to remove a `html.Div` wrapper for example that was combining a heading and a single component).

The new layout looks like this:

image

Non-visible changes:

- I moved the `*_content.py` files out of the top-level of the components page directory and into the relevant `components/<component_name>` folders. This means both that all the code specific to a particular component is in a single subdirectory, but also leaves the top level directory free for general purpose code such as the helpers, sidebar, and api documentation functions.


tcbegley:
Seems to have registered some of the moved files as deletions + creations. I'll mark this as wip for now to see if I can sort out the commit history

tcbegley:
I repeated the restructuring taking extra care and splitting the steps into multiple commits, but same thing happened again. Maybe my expectations of what should happen are simply wrong. Let me know if you think it's a problem.

pbugnion:
I think that whether a file is renamed or deleted/re-added is just a heuristic based on the number of lines that have changed. I don't think you can change that.

pbugnion:
Anyway, I think this looks great.

Issue #87, opened 2018-11-30
Documentation does not build on latest dash 


nadaouk:
Traceback (most recent call last):
File "run.py", line 3, in <module>
from app import App
File "/project/nadine/dash-bootstrap-components/docs/app.py", line 7, in <module>
from components_page import ComponentsPage
File "/project/nadine/dash-bootstrap-components/docs/components_page/__init__.py", line 1, in <module>
from .page import ComponentsPage  # noqa: F401
File "/project/nadine/dash-bootstrap-components/docs/components_page/page.py", line 5, in <module>
from .alerts_content import content as alerts_content
File "/project/nadine/dash-bootstrap-components/docs/components_page/alerts_content.py", line 3, in <module>
from .api_doc import ApiDoc
File "/project/nadine/dash-bootstrap-components/docs/components_page/api_doc.py", line 4, in <module>
from dash.development.base_component import (
ImportError: cannot import name 'filter_props'

tcbegley:
Thanks for reporting @nadaouk!

tcbegley:
Resolved by #88

PR #88, opened 2018-11-30
Apidoc fix

tcbegley:
This PR addresses #87.

The issue arose due to [this commit](https://github.com/plotly/dash/commit/4878245f2ead071bb5e78364be5c07c94e598b60#diff-7221a92fb7bc030889d57fabc119adfa) in the Dash repo that moved some of the internal functions we were relying on for the API documentation generation. I've fixed the import and also pinned the version of Dash required for the docs for good measure.

### DIFF
--- a/docs/components_page/api_doc.py
+++ b/docs/components_page/api_doc.py
@@ -1,7 +1,7 @@
 from itertools import chain
 
 import dash_html_components as html
-from dash.development.base_component import (
+from dash.development._py_components_generation import (
     filter_props,
     js_to_py_type,
     parse_events,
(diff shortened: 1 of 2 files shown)


pbugnion:
I think we should merge this since it obviously solves a problem we have right now. 

In the medium term, I'm a bit wary of depending on an internal component in Dash. I think it'd be worth vendoring this or finding a more stable source for it.

tcbegley:
Agreed, this isn't really satisfactory as a long term fix, finding a more stable source or vendoring in future sounds good.

PR #92, opened 2018-12-06
Add Table from DataFrame helper

tcbegley:
This PR adds a helper function for generating a `Table` component from a pandas DataFrame.

To maintain compatibility with older version of Python, I had to prevent `table.py` from being formatted by black. This is because black will insert a trailing comma after `**kwargs` in a function signature that is split onto multiple lines, but prior to 3.6 this would throw a SyntaxError.


srstevenson:
> To maintain compatibility with older version of Python, I had to prevent table.py from being formatted by black. This is because black will insert a trailing comma after **kwargs in a function signature that is split onto multiple lines, but prior to 3.6 this would throw a SyntaxError.

I would prefer to find a solution that doesn't necessitate excluding the entire file from formatting. What happens if you change `**table_kwargs` to `**kwargs`? The [Black documentation](https://github.com/ambv/black#trailing-commas) says:

> One exception to adding trailing commas is function signatures containing *, *args, or **kwargs. In this case a trailing comma is only safe to use on Python 3.6. Black will detect if your file is already 3.6+ only and use trailing commas in this situation. If you wonder how it knows, it looks for f-strings and existing use of trailing commas in function signatures that have stars in them.

srstevenson:
Also, I would suggest adding the relevant Trove classifiers to `setup.py` to document which versions of Python are supported.

tcbegley:
Thanks @srstevenson. I think the confusion my end was due to `table.py` being partially copied from elsewhere and originally containing an f-string which caused black to add the comma in the first place. It then won't subsequently remove it even if the file is no longer obviously safe for 3.6+ which lead me to overcorrect by ignoring the file completely when really I just needed to delete the comma. Anyway, should be good now.

Good shout on the classifiers, I've added that to this PR.

setup.py
```diff
@@ -32,6 +32,13 @@ def _get_long_description():
     author_email="engineering@asidatascience.com",
     url="https://github.com/ASIDataScience/dash-bootstrap-components",
     packages=find_packages(),
-    install_requires=["dash"],
+    install_requires=["dash", "pandas"],
```
pbugnion: I actually don't think we should rely on pandas at installation time for a helper function for a single component. If a user wants to use the `generate_table_from_df`, they already have pandas installed.

While there are fairly complex ways to specify optional dependencies, an easy route might just be to add a lin in the docs about how this requires pandas to be installed.

pbugnion:
This looks great! I made a comment about not including pandas as a dependency.

It might be worth adding API documentation for `generate_table_from_df`. Initially it's probably easiest to write that explicitly, rather than parse the docstring. I don't think we will have many functions in Python, so maintaining the documentation in 2 places might be acceptable.

dash_bootstrap_components/table.py
```diff
@@ -0,0 +1,85 @@
+import dash_bootstrap_components as dbc
+import dash_html_components as html
```
pbugnion: Does this introduce a depencency on `dash_html_components`?

tcbegley: Yes it does, good point. Should we put it in the dependencies?

setup.py
tcbegley: That's a fair point, I'll remove it.

dash_bootstrap_components/table.py
pbugnion: yes, I think we should (it's not like pandas because it's more lightweight and users of dbc are likely to need dash_html_components anyway, whereas they might not need pandas)

dash_bootstrap_components/table.py
```diff
@@ -0,0 +1,86 @@
...
+
+    if index:
+        df = df.reset_index()
+        if index_label is not None:
+            df = df.rename(columns={"index": index_label})
+
+    if header:
+        if isinstance(header, dict):
+            df = df.rename(columns=header)
+        elif isinstance(header, Sized):
```
tcbegley: @pbugnion I previously was checking if header was an instance of `(tuple, list, np.ndarray, pd.Index)`, but since we don't want `pandas` or `numpy` to be dependencies I need an alternative that can be used without importing either. This was the best I came up with, is there a better way do you think?

pbugnion: To clarify, I think it's fine to have pandas and numpy as dependencies for this part of the code. We could just nest the import inside the function so it's only imported when the function gets used, not when the module gets loaded.

My only concern was with explicitly specifying pandas as dependencies in `install_requires` in `setup.py`. That would force users to install pandas, whether they use `_generate_table_from_df` or not. I believe the most common way to specify optional dependencies is with `extras_require` (see [setuptools docs](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies)).

tcbegley: Ah right, I didn't think to put the imports inside the function, and was worried about pandas getting imported whenever you imported the module, that makes a lot more sense. Will amend!

tcbegley:
I think this is basically good to go now.

I've added API documentation for `from_dataframe` to the docs, modified the simple example to not use `pandas` (as it doesn't really make sense to give an example that uses a dataframe but doesn't call `from_dataframe`), and added `numpy` and `pandas` to `extras_require` in `setup.py`.

setup.py
```diff
@@ -32,6 +32,14 @@ def _get_long_description():
...
+    install_requires=["dash>=0.32.1", "dash-html-components"],
     include_package_data=True,
+    classifiers=[
+        "License :: OSI Approved :: Apache Software License",
+        "Programming Language :: Python :: 2.7",
+        "Programming Language :: Python :: 3.5",
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
+    ],
+    extras_require={"Table.from_dataframe": ["numpy", "pandas"]},
```
pbugnion: I wonder if the feature name should be something like `pandas-integration` or `pandas` or `dataframes`? We might imagine other functionality that maps from dataframes to components later?

pbugnion:
Thanks -- I think this is great! I added a single comment about the key in `extras_require`, but apart from that I think we should merge it.

setup.py
tcbegley: Yes, I was going to ask you about this actually as I thought my choice was a little clunky. I think `pandas` is nice and simple, but descriptive. I'll change to that.

dash_bootstrap_components/__init__.py
```diff
@@ -26,4 +29,6 @@
     _component._css_dist = _css_dist
 
 
-from ._components import *  # noqa
```
tcbegley: @pbugnion I think I accidentally isorted this to the top of the file, but I'm not totally confident what the significance of the placement here is. Everything appears to work for me, but I'm worried about issues that I might be blind to. Should I put it back at the bottom of the file?

PR #99, opened 2018-12-13
Update author email in setup.py

srstevenson:


pbugnion:
Thanks! Does this actually work, though? `engineering@` definitely goes to some people in the sherlock team. Unless `opensource@` has special handling, I expect it just goes to Robbie and Angie?

srstevenson:
Yes, it does: see my message in `#open-source` on Slack. `opensource@` is a Google Group that anyone in the ASI G Suite organisation can join, and which allows posting to from external domains.

pbugnion:
Ah I see -- I hadn't seen the Slack message. Thanks for setting that up!

PR #104, opened 2018-12-16
Explicitly import React in pure components

pbugnion:
Quite a lot of our components are so-called functional React components. They are defined as a function rather than as a class:

```js
const Alert = props => { /* do something */ }
```

While there is no explicit dependency on React, functional components still require having:

```js
import React from 'react'
```

... at the top.  This is because functional components desugar to explicit calls to `React.createComponent`, as explained [here](https://hackernoon.com/why-import-react-from-react-in-a-functional-component-657aed821f7a).

So far, this hasn't actually caused us any issues because:

- in the demo, we explicitly imported react and react-dom in `<script />` tags. This injects `React` into the global namespace, so the components still worked.
- in the actual generated components, Dash intentionally requires React to be injected from the outside so it doesn't matter whether we explicitly set it in components. To avoid bundling React ourselves (which would duplicate it, since Dash already provides it), we declare it as an external dependency in `webpack.config.dist.js`. This tells webpack to just assume the keyword `React` exists in the global namespace.

This needs to be fixed to allow embedding *dash-bootstrap-components* into third party JS libraries. For this, we need to distribute *dash-bootstrap-components* as a UMD distribution. We will probably want to declare react and react-dom as external libraries ([reactstrap do this too](https://github.com/reactstrap/reactstrap/blob/master/webpack.base.config.js#L42)). For this to work, the imports need to be correct.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -6,8 +6,6 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/css/bootstrap.min.css" integrity="sha384-MCw98/SFnGE8fJT3GXwEOngsV7Zt27NXFoaoApmYm81iuXoPkFOJwJ8ERdknLPMO" crossorigin="anonymous">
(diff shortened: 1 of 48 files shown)


tcbegley:
Thanks for fixing this and for taking the time to explain in such detail!

pbugnion:
Thanks! I'm probably going to leave this open until I have the full pipeline to push to NPM.

PR #105, opened 2018-12-16
Allow publishing to npm

pbugnion:
This PR adds publishing to NPM as part of the build chain. I've made a small repo showing how to use the built artefacts: https://github.com/pbugnion/dash-bootstrap-components-examples . This is released as 0.2.6-alpha1.

The distribution pushed to NPM is different from the distribution used by dash-renderer to build the components when dbc is used through Python. This is because dash-renderer expects its modules to be available on the `window` object (e.g. dash-bootstrap-components is available at `window["dash_bootstrap_components/_components"]`). However, when using a dash-bootstrap-components as a library in other JS components, the library needs to be made available through other import mechanisms (typically [CommonJS](https://en.wikipedia.org/wiki/CommonJS), so basically `import` and `export` declarations work).

Practically, this translates into two separate `output` sections for the webpack configuration. The bundle used by dash-renderer is created with `libraryTarget: window`. The bundle used by other libraries is created with `libraryTarget: umd` (UMD is a superset of CommonJS). For the new UMD definition, I mostly copied Reactstrap's configuration.

To work, this requires PR #104.


tasks.py
```diff
@@ -111,6 +111,7 @@ def build_publish(version):
     info("Building JavaScript components")
     build_js()
     info("Building and uploading Python source distribution")
+    info("Pypi credentials:")
```
srstevenson: `s/Pypi/PyPI/`

pbugnion: good call, thanks

pbugnion:
Actually, before I merge this -- @tcbegley it'd be great to check that you can still create releases. We will need to do the following, I think:
- You should create an account on NPM if you don't have one already
- I presumably can then add you
- You need to run `npm login` on any computer that you plan to push to NPM from. This writes an `.npmrc` file in your home directory that has an authorisation token. (on SherlockML, you only need to do this once, not once per server, since your home directory is persistent).
- Try pushing `0.2.6-alpha2`.

tcbegley:
Sorry for the delay, I've successfully published to npm without issue. Happy for you to merge when ready.

pbugnion:
Great! Will merge this and do a release tomorrow.

Issue #74, opened 2018-11-23
Document Input components


pbugnion:

tcbegley:
Resolved by #108

PR #108, opened 2018-12-22
Add input components documentation

tcbegley:
This PR adds documentation for the input components `Input`, `RadioItems` and `Checklist`, addressing #74. 

It also adds documentation for `FormFeedback` to the forms docs. Originally I added that as part of the docs for `Input`, but I felt it made more sense to keep it with the docs for the other `Form*` components.

Documentation for `RadioItems` and `Checklist` has been moved from the forms page to the input page and been expanded to include the inline styling.

I decided against documenting the datepickers for now as it's a little hacky and only really works for default Bootstrap. I want to fix it to work properly with any Bootstrap theme.


pbugnion:
This is great, thank you!

PR #109, opened 2018-12-22
Clean examples directory

tcbegley:
This PR is intended as a first step to improving the examples directory. I have

* Removed the Bootstrap CSS sheet in favour of using `dbc.themes`. The examples will still work out of the box if people clone the repo.
* Removed example apps that had primarily sprung up as tests during development and weren't great showcases / examples for users.
* Renamed `app.py` to `components.py` and added new components. This app serves as a simple demo of all the available components.
* Added the `shiny` directory. This contains ports of the simple examples from the [shiny gallery](https://shiny.rstudio.com/gallery/). I will probably add some of the more complex ones in the future too. I think it could be useful for anyone coming from Shiny to see how the same sorts of layouts can be built with `dash-bootstrap-components`.

I'd like to add many more examples, but I think it makes sense to leave those for future PRs.


pbugnion:
This is great. I have a couple of suggestions:

- should we rename the directory `gallery` instead of `shiny`? The fact that these are ported from RShiny is a bit irrelevant (though we should probably acknowledge RShiny somewhere).
- should we host the data files on a remote server (e.g. S3) and load them dynamically? That way, they don't have to be committed to the repo. If we think that's useful, I can create an S3 bucket in e.g. the ASI account and we can upload the data there. Committing large files is generally considered poor practice since it leads to large diffs and slows down every git command. If we think storing the data on S3 is the way to go, I'm happy to set that up and push directly onto this branch.

examples/components.py
```diff
@@ -4,17 +4,14 @@
...
-
+app = dash.Dash(external_stylesheets=[dbc.themes.BOOTSTRAP])
 
 navbar = dbc.Navbar(
-    brand="Dash Bootstrap components",
-    brand_href=DBC_GITHUB,
-    sticky="top",
     children=[
-        dbc.NavItem(dbc.NavLink("ASI", href="https://www.asidatascience.com")),
+        dbc.NavItem(dbc.NavLink("Github", href=DBC_GITHUB)),
```
srstevenson: s/Github/GitHub/

tcbegley:
* That's fair. In the wordcloud and faithful examples I gave links to the shiny analogues in module docstrings at the top of the `app.py` files. Perhaps adding something similar to the other two examples will suffice and then we rename the directory as you suggest.
* I think hosting data on S3 is a great idea. Happy for you to set that up and work on this branch if you'd like to take a stab at it.

pbugnion:
I've created a CDN backed by S3 at `cdn.opensource.faculty.ai`. I seem to be running into some redirection issues at the moment (because of [this](https://forums.aws.amazon.com/thread.jspa?threadID=216814)), but hopefully it'll stabilise in the next couple of hours. When it does, I'll update the examples in this PR to avoid having the data in this repo.

Data files will be available on, e.g. `https://cdn.opensource.faculty.ai/wordcloud/romeo-and-juliet.txt`. 

I chose to create a single CDN for open source at faculty. Data files are likely to be useful across multiple repositories. I created the CDN under the faculty name directly because doing it once is painful enough.

@srstevenson you may be interested in this too.

pbugnion:
Tom -- if you don't have access to ASI's root AWS account we can add you.

pbugnion:
I opened PR #115, which targets this PR.

tcbegley:
> Tom -- if you don't have access to ASI's root AWS account we can add you.

I don't think I do. At the very least looking at the S3 console currently I don't see anything that's obviously what you've set up for this.

PR #114, opened 2018-12-24
Add task to push documentation to Heroku

pbugnion:
This adds a `documentation` task that pushes current master to Heroku. You can now just run:

```
inv documentation
```

I tested this when upgrading the documentation to 0.2.6.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -92,7 +92,7 @@ def documentation(ctx):
     help={
         "version": "Version number to finalize. Must be "
         "the same version number that was used in the release."
-    }
+    },
 )
 def postrelease(ctx, version):
     """
@@ -173,7 +173,8 @@ def set_documentation_version(version):
         docs_requirements = f.readlines()
     for iline, line in enumerate(docs_requirements):
         if "dash_bootstrap_components" in line:
-            docs_requirements[iline] = f"dash_bootstrap_components=={version}\n"
+            updated_line = f"dash_bootstrap_components=={version}\n"
+            docs_requirements[iline] = updated_line
     with open(docs_requirements_path, "w") as f:
         f.writelines(docs_requirements)
 
--- a/tasks.py
+++ b/tasks.py
@@ -102,6 +102,15 @@ def postrelease(ctx, version):
     run("git push origin master")
 
 
+@task
+def documentation(ctx):
+    """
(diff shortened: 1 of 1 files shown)


pbugnion:
> Would it be worth allowing updating the docs requirements with this?

Would your expectation be that this would edit `docs/requirements.txt` file? If so, happy to add that in. An alternative might be to:
- bundle bumping the version in `docs/requirements.txt` as part of `inv release`. It probably makes sense to update the documentation version as part of a release.
- bundle pushing to heroku as part of `inv postrelease`. Once we've verified that a release is correct, we probably always want to update the documentation.

tcbegley:
> Would your expectation be that this would edit `docs/requirements.txt` file?

Yes, that's what I had in mind.

I think bundling the version bump for `docs/requirements.txt` into `inv release` makes sense, I can't really think of a scenario in which we wouldn't want to run the docs on the lastest version.

While I don't have a problem with pushing to heroku as part of `inv postrelease`, did you have in mind that this would replace `inv documentation`? I think there are cases where we would want to update the docs without making a new release of the python package, so it might be useful to have `inv documentation` available in such cases.

pbugnion:
> I think bundling the version bump for docs/requirements.txt into inv release makes sense, I can't really think of a scenario in which we wouldn't want to run the docs on the lastest version.

Great! I can set that up.

> While I don't have a problem with pushing to heroku as part of `inv postrelease`, did you have in mind that this would replace `inv documentation`? I think there are cases where we would want to update the docs without making a new release of the python package, so it might be useful to have `inv documentation` available in such cases.

Good question. I was probably thinking of keeping `inv documentation` as a stand-alone task and declare a dependency on it in `inv postrelease`. That would deploy documentation automatically after a release, but it would also let you deploy doc changes automatically.

pbugnion:
I think this is now ready for review again. To summarise, this PR:
- bumps the version of dbc in `docs/requirements.txt` as part of `invoke release`.
- pushes documentation as part of `invoke postrelease`
- adds a stand-alone `invoke documentation` task that pushes the current master branch to heroku.

tcbegley:
This looks great to me. Happy for you to merge and test.

pbugnion:
Great! Will do that tomorrow (Sunday).

Issue #101, opened 2018-12-13
Support custom NavbarBrand instead of just a string?


mkhorton:
Hi, I'm not sure if I'm missing an obvious solution to this, but it seems like as currently written the Navbar brand has to be a string, even though the NavbarBrand component itself could in principle contain other items as children (eg an Html.Img)?

I'm looking for an easy way to add a left-aligned logo to a Navbar instead of only brand text.

Many thanks for releasing this component suite, it looks very useful.

mkhorton:
Actually may be easier to do this via css and the brand_style.

tcbegley:
Hey @mkhorton, you're right, the brand does have to be a string. 

Currently it's only possible to pass Dash components as the children of other Dash components, not any other prop. So in particular you can't pass, e.g. `html.Img` to `brand`. There was [an issue](https://github.com/plotly/dash-renderer/pull/32) opened a while back on the `dash-renderer` repo that would have addressed this, but it's a big change and has been dormant for a little while, so I don't expect it to happen any time soon.

This restriction, and also because we were trying to keep the interface of `Navbar` relatively simple, meant that we sacrificed some customisability of the `Navbar` component.

I posted on the [plotly community forums](https://community.plot.ly/t/show-and-tell-dash-bootstrap-components/16614/15?u=tcbegley) about how you could more manually construct a Bootstrap style navbar using `dash-html-components` and applying the relevant Bootstrap classes with `className`. Perhaps that's of use to you for now.

As a longer-term fix, since multiple people are asking similar questions, perhaps we can create a `CustomNavbar` component that gives you full control over the contents, but would require some careful construction of the `children`. @pbugnion what do you think?

pbugnion:
I agree with the idea of a new custom component that gives more control over the children. It seems like a good way to give the user (almost) arbitrary flexibility, but still keeping a simple navbar for people who don't need all that flexibility.

ned2:
I would love a custom component that supports more control over the children. I just came here to open an issue asking about just this.

In addition to the limitation of only having strings as the `brand` prop, I was also hoping to be able to add Dash components to parts of the Navbar that are not nav-items. For example, including an form element before the nav items, as in [this Bootstrap dashboard](https://getbootstrap.com/docs/4.1/examples/dashboard).

tcbegley:
[...]

PR #118, opened 2018-12-28
Custom navbar

tcbegley:
This PR adds a `CustomNavbar` component giving users full control over the children, addressing #101


src/components/nav/CustomNavbar.js
```diff
@@ -0,0 +1,80 @@
...
+   * the page.
+   */
+  sticky: PropTypes.string,
+
+  /**
+   * Sets the color of the Navbar, options: primary, secondary, success, warning, danger, info, light.
+   */
+  color: PropTypes.string,
+
+  role: PropTypes.string,
```
pbugnion: Should we document the `role`? AFAIK, this is used for telling the browser what the function of an element is (see e.g. [here](https://stackoverflow.com/questions/10403138/what-is-the-purpose-of-the-role-attribute-in-html)).

pbugnion:
This is great, thanks!

Should we:
1. put a link to the `advanced-components-usage/CustomNavbar.py` example in the markdown used to document navbars? Otherwise, the examples are not very discoverable, which is a shame because they're great.
2. consider whether `CustomNavbar` is really the best name? We could imagine renaming `Navbar` ->  `SimpleNavbar` (or `NavbarSimple` for discoverability) and have `CustomNavbar` as `Navbar`, since it's closer to the original Bootstrap navbar. Alternatively, we could rename `CustomNavbar` to `RawNavbar` (or `NavbarRaw`) or `BasicNavbar`?

docs/components_page/components/navbar/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+def get_content(app):
+    return [
+        html.H2("Navbar"),
+        html.P(
+            dcc.Markdown(
+                "There are two navbar components in "
+                "`dash-bootstrap-components`, `Navbar` and `CustomNavbar`. "
+                "`Navbar` is simpler but less flexible, whereas "
+                "`CustomNavbar` is fully customisable, but requires more "
+                "boilerplate to get working."
```
pbugnion: This is great!

src/components/nav/CustomNavbar.js
tcbegley: Yes, I'll add this.

tcbegley:
Thanks for reviewing!

1. Yes, good idea, I'll add that.
2. Agreed, I think that reserving `Navbar` for the fully customisable option to more closely shadow Bootstrap is probably preferable. The main reason I opted for `CustomNavbar` was to preserve backwards compatibility. If we decide to break that, what's the best way we can warn users who have already used `Navbar` in their code that the usage will change if they upgrade?

src/components/nav/CustomNavbar.js
tcbegley: The documentation in `dash-html-components` for this is fairly minimal. see e.g. [here](https://github.com/plotly/dash-html-components/blob/master/src/components/Button.react.js#L63-L66)

Do you think we need any more detail than that or shall I opt for the same sort of thing?

src/components/nav/CustomNavbar.js
```diff
@@ -0,0 +1,80 @@
...
+
+const CustomNavbar = props => {
+  const {children, ...otherProps} = props;
+
+  return <RSNavbar {...otherProps}>{children}</RSNavbar>;
+};
+
+CustomNavbar.defaultProps = {
+  fluid: false,
+  color: 'light',
```
tcbegley: There's a bit of an issue here (and also with `Navbar`) I think with how the background colour is set. The `color` prop is used by reactstrap to add `bg-${color}` to the classes. These classes use the `!important` rule to set the background colour which means to set the background colour to something custom using either CSS classes or inline style arguments you need to pass something like `color=False` to prevent `bg-light` being applied by default.

I think there are two options:
1. Remove the default argument, forcing people to explicitly state a colour rather than relying on defaults.
2. Allow a hex code to be passed to `color`. We would check the `color` prop before passing it on to the underlying reactstrap component or using it to set the `background-color` style argument as appropriate.

Of course these are not mutually exclusive and could also be combined.

@pbugnion what do you think?

tcbegley: I've implemented option 2. for now, it works quite well I think, and in case people really want to do something very custom they always have the option of passing `color=False` to switch everything off entirely.

docs/components_page/components/navbar/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+from ...api_doc import ApiDoc
+from ...helpers import (
+    ExampleContainer,
+    HighlightedSource,
+    load_source_with_environment,
+)
+from ...metadata import get_component_metadata
+from .navbar import navbar
+
+GITHUB_EXAMPLES = "https://github.com/ASIDataScience/dash-bootstrap-components/blob/master/examples/advanced-component-usage/CustomNavbar.py"  # noqa
```
tcbegley: I changed this to link directly to the example. As it stands, this link will break because the file isn't there yet. Once this is merged into master it should all be fine 🤞

src/components/nav/CustomNavbar.js
pbugnion: I think option 2 is the right call.

src/components/nav/Navbar.js
```diff
@@ -144,7 +163,11 @@ Navbar.propTypes = {
   sticky: PropTypes.string,
 
   /**
-   * Sets the color of the Navbar, options: primary, secondary, success, warning, danger, info, light.
+   * Sets the color of the Navbar. Main options are primary, light and dark, default light.
+   *
+   * You can also choose one of the other contextual classes provided by Bootstrap
+   * (secondary, success, warning, danger, info, white) or any color of your choice
+   * specified either by its HTML color name or by its hex code.
```
pbugnion: Why not *or any valid CSS color of your choice (e.g. a hex code, a decimal code or a CSS color name)* ? AFAICT, this allows any valid CSS color (apart from `white`), so `rgba(1, 2, 3, 0.2)` is valid?

src/components/nav/CustomNavbar.js
pbugnion: I think copying the dash-html-components string is the right call.

pbugnion:
> The main reason I opted for CustomNavbar was to preserve backwards compatibility.

We are still at a stage where we can break backwards compatibility without too many consequences, I think.

> what's the best way we can warn users who have already used Navbar in their code that the usage will change if they upgrade?

I think that, if we were more mature, the correct way would be to introduce a deprecation cycle like:

1. rename `Navbar` to `CustomNavbar`. Have `Navbar` just proxy to `CustomNavbar`, but raising a deprecation warning when you try to use it.
2. wait some length of time proportional to how mature the project is.
3. replace existing `Navbar` with the new code and bump the version number.

This gives people a chance to migrate their code. In the interest of moving forward, a more pragmatic solution might be to do a cursory check of libraries that we think use dash-bootstrap-components to see if they use `Navbar`.

src/components/nav/Navbar.js
tcbegley: Good call, added.

tcbegley:
I think this is ready for a final (🤞) review.

I've refactored `Navbar` -> `NavbarSimple` and `CustomNavbar` -> `Navbar`, and made the relevant changes to the example, tests and docs.

I've also made the suggested changes to the docstring for `color`.

docs/components_page/components/navbar/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+navbar_source = (HERE / "navbar.py").read_text()
+
+
+def get_content(app):
+    return [
+        html.H2("Navbar"),
+        html.P(
+            dcc.Markdown(
+                "There are two navbar components in "
+                "*dash-bootstrap-components*, `NavbarSimple` and `Navbar`. "
```
pbugnion: `Navbar`. `NavbarSimple` renders in a somewhat confusing way:

screenshot 2019-01-06 at 12 32 56

Maybe: "The `NavbarSimple` component easier to use but less flexible"?

pbugnion:
I think this is great -- made one comment inline but, apart from that, we should merge this.

docs/components_page/components/navbar/__init__.py
tcbegley: Good call, I'll change this then merge.

tcbegley:
@pbugnion given the breaking change to the `Navbar` component shall I release this as `0.3.0` rather than `0.2.9`?

PR #121, opened 2019-01-07
Bump bootstrap version to 4.2.1

tcbegley:
This PR updates the CDN links in the `themes` module to use Bootstrap version 4.2.1.

Though we can't yet take advantage of the new components, there are some new utility CSS classes available which I think are useful to have access to.

Here's a link to the [bootstrap release notes](https://github.com/twbs/bootstrap/releases/tag/v4.2.0) for reference (for version 4.2.0, it was bumped to 4.2.1 just to republish on npm)


pbugnion:
Should we update the bootstrap links in the documentation as well? I believe there is one on the index page.

tcbegley:
Good spot! I'll update then merge.

Issue #124, opened 2019-01-12
Strange behavior with new Navbar


villoro:
Hello @tcbegley 

This is the problem I was telling you in my repo.

With the new `navBar` component introduced in version `0.3.0` there is a strange behavior when parameter 'nav' is set to True. It adds a point and make the navbar bigger than expected

This is **without** `nav=True`:
![image1](https://user-images.githubusercontent.com/10529654/51076074-344dde00-1694-11e9-9250-7e2573ebadfa.PNG)

This is with `nav=True`:
![image2](https://user-images.githubusercontent.com/10529654/51076076-37e16500-1694-11e9-911d-a498e9a4daf5.PNG)

I would like to set `nav=True` so that the dropdown does not appear as a button. Something similar to the `navbarSimple` example in the [doc](https://dash-bootstrap-components.opensource.asidatascience.com/l/components/navbar).

And here a full example:
```python
from dash import Dash
from dash_bootstrap_components.themes import BOOTSTRAP

import dash_bootstrap_components as dbc
import dash_html_components as html

app = Dash('expensor_personal', external_stylesheets=[BOOTSTRAP])

navbar = dbc.Navbar(
    [
        dbc.Row(
            [
                dbc.Col(dbc.NavbarBrand("Dash", className="ml-2")),
            ],
            align="center",
            no_gutters=True,
        ),
        dbc.Row(
            [
                dbc.DropdownMenu(
                    label="Pages",
                    children=[
                        dbc.DropdownMenuItem("Page 1", href="/"),
                        dbc.DropdownMenuItem("Page 2", href="/"),
                    ],
                    className="mr-1",
                    nav=True
                ),
                dbc.Button(
                    "Button1",
                    className="mr-1",
                ),
            ],
            no_gutters=True,
            className="ml-auto",
            align="center",
        )
    ],
)

app.layout = html.Div([navbar])

if __name__ == '__main__':
    app.run_server(debug=True)
```

PR #125, opened 2019-01-15
Document Nav

tcbegley:
This PR adds documentation for the `Nav` component, primarily motivated by confusion documented in #124. In that issue I outlined the following tasks:

- [x] ~Clarify docstring of `DropdownMenu`~ *(no change here, I decided they were explicit enough)*
- [x] Add page to docs for `Nav`
- [x] Mention `Nav` explicitly in the docs for `Navbar`

In the course of preparing the documentation, guided by both Bootstrap and reactstrap docs, I also did the following:

- Add `external_link` prop to `NavLink` which gets passed to the underlying `Link` component. Previously it was not possible to set this in the Python interface.
- Add `n_clicks` and `n_clicks_timestamp` props to `NavLink` to allow usage as a button in callbacks. Implementation is following that of `DropdownMenuItem`.
- Add `card` prop to `Tabs` to allow use of `Tabs` inside a `CardHeader`, This has also been added to the card documentation.
- Previously the `id`, `className` and `style` props of `Tabs` didn't actually get used by the underlying React component (`id` could still be used correctly in callbacks), I have passed all of these to the `Nav` component that `Tabs` is built from.
- Experimented with a display heading and a lead paragraph on the docs pages for `Nav` and `Tabs`, inspired by the official Bootstrap documentation. If we decide we like it we can roll this out to all the other component pages too in future.

I made prerelease `0.3.1rc1` for testing all of the above.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ dash_bootstrap_components/metadata.json
 *.pyc
 dist
 changelog.tmp
+dash_bootstrap_components/_components/
--- a/docs/components_page/components/nav/__init__.py
+++ b/docs/components_page/components/nav/__init__.py
@@ -107,6 +107,14 @@ def get_content(app):
         ),
         ExampleContainer(nav_pill),
         HighlightedSource(nav_pill_source),
+        html.H4("Tabs"),
+        html.P(
+            dcc.Markdown(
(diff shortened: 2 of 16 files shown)


pbugnion:
This looks great! Thanks for putting it together..

PR #127, opened 2019-01-29
Remove Events

tcbegley:
This PR removes events from *dash-bootstrap-components* as they are [no longer supported by Dash](https://github.com/plotly/dash/issues/531).

### DIFF
--- a/docs/components_page/api_doc.py
+++ b/docs/components_page/api_doc.py
@@ -1,10 +1,7 @@
-from itertools import chain
-
 import dash_html_components as html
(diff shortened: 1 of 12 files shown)


pbugnion:
This looks good -- thank you!

Are there going to be new props on some of those widgets to replace events? For instance, on a link, how is a consumer going to know that a link was clicked? Through the `n_clicks` or `n_clicks_timestamp` prop?

pbugnion:
That's great. Thank you!

PR #128, opened 2019-01-29
Add key prop

tcbegley:
This PR adds a key prop to all components. It has been rebased on top of #127


pbugnion:
This looks great!

PR #129, opened 2019-01-31
Improve docstrings

tcbegley:
This PR expands and makes consistent many of the existing docstrings and adds documentation to props that were previously undocumented.

While working on this I realised that the `FormFeedback` example in the docs was overcomplicated, and simplified it. I also noticed that the `disabled` prop of `FormGroup`, the `tabs` prop of `Nav` and `inline` prop of `FormText` serve no useful purpose so I removed them.

### DIFF
--- a/docs/components_page/components/form/__init__.py
+++ b/docs/components_page/components/form/__init__.py
@@ -57,7 +57,11 @@ def get_content(app):
                 "set the `valid` property. The feedback will display if its "
                 "validity matches the validity of the `Input`. For this to "
                 "work you must wrap your `Input` and `FormFeedback` in a "
-                "`FormGroup` component."
+                "`FormGroup` component. In the below example we create two "
+                "`FormFeedback` components, one valid, one not valid. "
+                "They will appear / disappear automatically as the `valid` "
+                "and `invalid` properties of `Input` are set to `True` / "
+                "`False` respectively."
             )
         ),
         ExampleContainer(
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -82,7 +82,8 @@ Button.propTypes = {
   active: PropTypes.bool,
 
   /**
-   * Create block button.
+   * Create block level button, one that spans the full width of its parent.
+   * Default: False
    */
   block: PropTypes.bool,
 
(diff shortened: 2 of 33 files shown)


docs/components_page/components/form/__init__.py
```diff
@@ -57,7 +57,11 @@ def get_content(app):
                 "set the `valid` property. The feedback will display if its "
                 "validity matches the validity of the `Input`. For this to "
                 "work you must wrap your `Input` and `FormFeedback` in a "
-                "`FormGroup` component."
+                "`FormGroup` component. In the below example we create two "
+                "`FormFeedback` components, one valid, one not valid."
```
pbugnion: Missing a space after `valid.`.

src/components/Button.js
```diff
@@ -82,7 +82,7 @@ Button.propTypes = {
   active: PropTypes.bool,
 
   /**
-   * Create block button.
+   * Create block level button, one that spans the full width of its parent.
```
pbugnion: Is there value in adding "(default False)"?

src/components/card/Card.js
```diff
@@ -39,22 +39,23 @@ Card.propTypes = {
...
    * Card color, options: primary, secondary, success, info, warning, danger,
-   * link. Default is secondary
+   * light, dark. Default is light.
    */
   color: PropTypes.string,
 
   /**
-   * Wrap all children in a CardBody element (basically add some padding)
+   * Apply the `card-body` class to the card, so that there is no need to also
+   * include a CardBody component in the children of this Card.
```
pbugnion: Is there value in saying: "(default False)"?

src/components/input/Input.js
```diff
@@ -165,20 +165,71 @@ Input.propTypes = {
...
 
+  /**
+   * The value of the Input
+   */
   value: PropTypes.string,
+
+  /**
+   * The initial size of the control. This value is in pixels unless the value
+   * of the type attribute is text or password, in which case it is an integer
+   * number of characters. Starting in, this attribute applies only when the
```
pbugnion: That sentence is a bit strange. Starting in what?

src/components/input/Input.js
```diff
@@ -165,20 +165,71 @@ Input.propTypes = {
...
+   * type attribute is set to text, search, tel, url, email, or password,
+   * otherwise it is ignored. In addition, the size must be greater than zero.
+   * If you do not specify a size, a default value of 20 is used.' simply
```
pbugnion: It looks like something's gone wrong with `simply` here?

src/components/input/Input.js
tcbegley: Whoops, my bad. I copied most of these from the [dash-core-components Input](https://github.com/plotly/dash-core-components/blob/master/src/components/Input.react.js) and didn't check them that carefully. You're right this doesn't make a whole lot of sense.

PR #132, opened 2019-02-08
Switch domain names to faculty.ai

pbugnion:
The documentation is now live at https://dash-bootstrap-components.opensource.faculty.ai/. This PR switches links to the new URL structure. 

At the moment, both `dash-bootstrap-components.opensource.{asidatascience.com,faculty.ai}` point to the Heroku app. When this is merged, I'll set up redirection from `...asidatascience.com` to `faculty.ai` so the former gets progressively purged from people's browsers.


tcbegley:
This is great, thanks for taking care of it!

PR #143, opened 2019-03-06
Add option to right align the dropdown menu

jmsblah:
Change to add option to align the dropdown menu along the right side of its parent. Default: False.

image

image

### DIFF
--- a/docs/components_page/components/dropdown/__init__.py
+++ b/docs/components_page/components/dropdown/__init__.py
@@ -5,10 +5,10 @@ import dash_html_components as html
 from ...api_doc import ApiDoc
 from ...helpers import ExampleContainer, HighlightedSource
 from ...metadata import get_component_metadata
+from .alignment import dropdown as dropdown_alignment
+from .direction import dropdown as dropdown_direction
 from .simple import dropdown as dropdown_simple
 from .size import dropdown as dropdown_size
-from .direction import dropdown as dropdown_direction
-from .alignment import dropdown as dropdown_alignment
 
 HERE = Path(__file__).parent
 
--- a/docs/components_page/components/dropdown/__init__.py
+++ b/docs/components_page/components/dropdown/__init__.py
@@ -5,15 +5,17 @@ import dash_html_components as html
 from ...api_doc import ApiDoc
 from ...helpers import ExampleContainer, HighlightedSource
 from ...metadata import get_component_metadata
-from .direction import dropdown as dropdown_direction
 from .simple import dropdown as dropdown_simple
 from .size import dropdown as dropdown_size
+from .direction import dropdown as dropdown_direction
+from .alignment import dropdown as dropdown_alignment
 
 HERE = Path(__file__).parent
 
 dropdown_simple_source = (HERE / "simple.py").read_text()
 dropdown_size_source = (HERE / "size.py").read_text()
 dropdown_direction_source = (HERE / "direction.py").read_text()
+dropdown_alignment_source = (HERE / "alignment.py").read_text()
 
 content = [
     html.H2("DropdownMenu"),
@@ -25,6 +27,9 @@ content = [
     html.H4("DropdownMenu direction"),
     ExampleContainer(dropdown_direction),
     HighlightedSource(dropdown_direction_source),
+    html.H4("DropdownMenu alignment"),
+    ExampleContainer(dropdown_alignment),
+    HighlightedSource(dropdown_alignment_source),
     ApiDoc(
         get_component_metadata("src/components/DropdownMenu.js"),
         component_name="DropdownMenu",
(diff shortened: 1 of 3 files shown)


tcbegley:
Thanks for preparing this @jmsblah, it looks great!

Here's a minimal example that demonstrates the sizing issues I mentioned

```python
import dash

app.layout = dbc.Container(
    dbc.DropdownMenu(
        [dbc.DropdownMenuItem("Item 1"), dbc.DropdownMenuItem("Item 2")],
        right=True,
        label="Open",
    )
)


if __name__ == "__main__":
    app.run_server()
```

The container here is just to add a margin so that there is enough space for right alignment to work, you don't actually need it to see the sizing issues.

If you run that and open the dropdown, you see something like this:

image

The dropdown takes the full width of the parent container (in your case, testing inside a Navbar you wouldn't have seen this because the parent container is narrower than the minimum width of the dropdown).

If you run the exact same code with the Dash v1 pre-releases (`pip install dash==1.0.0a1 dash-renderer==1.0.0a2 dash-html-components==1.0.0a2 dash-core-components==1.0.0a4`, I recommend you do this in a virtual environment or something) you see this

image

I am pretty sure the reason for this is related to the discussion [here](https://github.com/facultyai/dash-bootstrap-components/issues/62#issuecomment-439684155). Consequently I think the best thing is if we merge this into the `dash-v1` branch and it will get released when Dash v1 does.

Before we do that though, it would be great if we added some documentation for this to the [DropdownMenu documentation](https://dash-bootstrap-components.opensource.faculty.ai/l/components/dropdown_menu). If you want to have a go at that, take a look [here](https://github.com/facultyai/dash-bootstrap-components/tree/master/docs/components_page/components/dropdown). The page content is specified in `__init__.py` and each example is contained in a separate file in the same directory.

jmsblah:
@tcbegley Nice! I have a virtual environment for this change, will test it out again to reproduce and test against Dash v1 just to see the behavior. It will be great for training. Will take a look at the documentation as well, probably in the next few days.

What do I need to do to move this pull request into `dash-v1` branch?

tcbegley:
> What do I need to do to move this pull request into `dash-v1` branch?

Click on the edit button at the top right of this page, then select the `dash-v1` branch from the dropdown menu. Turns out I have the permissions to do it, but I'll let you do it so you can see where it is :)

image

jmsblah:
So, I replicated the sizing issue in the current dash version and I was also able to see it fixed when using dash v1 pre-releases.

On the documentation, I don't understand how the files are rendered into a webpage but I think I was able to mimic the current examples to generate a new one referring to right-alignment.

To be consistent with the Bootstrap documentation, I opted to create a new section for the right alignment feature. It looks like the only thing I needed to do was to create a new py file and reference it in __init__.py. It seems to me that the **Keywords Arguments** section will be auto generated but the js file.

Let me know your thoughts, also, if you can give point me to any documentation that explains how these examples are rendered, that will be great. 

image
.

image

tcbegley:
This is great! Thank you for taking care of that.

> I don't understand how the files are rendered into a webpage

Sorry, I should have said. The docs are basically just a big Flask / Dash app. You can run it with `python docs/run.py`. Just make sure you've installed the version of *dash-bootstrap-components* with the relevant changes. Then you just navigate to the page you changed and you can see all the examples running. Your addition looks perfect.

image

jmsblah:
Cool! I can try to contribute more. Let me know if there is something else I can go slowly working on.

tcbegley:
I think the lowest hanging fruit at the moment is documentation and examples. It's maybe a bit less exciting, but any contributions there are very welcome. Ultimately we'd like to have the documentation for each page be like [the page for navs](https://dash-bootstrap-components.opensource.faculty.ai/l/components/nav), with text explanations instead of just examples, and a bit of an introduction at the top. With those more complete pages in the docs I've been taking my cues from the corresponding pages in the [Bootstrap docs](https://getbootstrap.com/docs/4.3/components/navs/) and the [reactstrap docs](https://reactstrap.github.io/components/navs/).

In terms of components, there's [breadcrumbs](https://reactstrap.github.io/components/breadcrumbs/) and [pagination](https://reactstrap.github.io/components/pagination/) that I would like to add. The former would probably be a good one, I don't think it would be too hard to make a component for that, the latter will probably take some discussion about the best way to implement it in a user-friendly way.

Also a new component in Bootstrap as of a couple months ago is [toasts](https://getbootstrap.com/docs/4.3/components/toasts/). These haven't made it into reactstrap yet, but there's an open pull request for that on the [reactstrap repo](https://github.com/reactstrap/reactstrap/pulls). Once it's incorporated there, you could add a `Toast` component. That would be another addition that has to wait for Dash v1 though.

tcbegley:
This has been merged and is available in version 0.5.0 🎉

PR #145, opened 2019-03-08
Fix unsorted imports in __init__.py files

pbugnion:
Since isort 4.3.5 ([release notes](https://github.com/timothycrosley/isort/releases/tag/4.3.5)), `__init__.py` files are also sorted. I'm not sure why isort did not consider these files before.

This means that any `__init__.py` file that was previously unsorted breaks our build pipeline.


tcbegley:
> I'm not sure why isort did not consider these files before.

Agreed, that seems weird. Happy for you to merge this.

PR #146, opened 2019-03-08
Add modals

tcbegley:
This PR adds modals to *dash-bootstrap-components* along with reasonably extensive documentation.

This should be merged into the `dash-v1` branch as it requires `dash-renderer` to be using React 16.

There is one outstanding thing to change before merging that I've commented on inline.


src/components/modal/Modal.js
```diff
@@ -0,0 +1,149 @@
...
+    if (nextProps.is_open != prevState.modalOpen) {
+      return {modalOpen: nextProps.is_open};
+    } else return null;
+  }
+
+  render() {
+    const {children, className, scrollable, ...otherProps} = this.props;
+    const classes = classnames(className, {
+      'modal-dialog-scrollable': scrollable
+    });
```
tcbegley: This is a temporary work-around, a `scrollable` prop for `Modal` will make its way into a future release of reactstrap, see [here](https://github.com/reactstrap/reactstrap/pull/1435). When it does, this should be removed in favour of passing the `scrollable` prop down to reactstrap.

Issue #151, opened 2019-03-13
Style is not set on the Tab component


mikesmith1611:
The style property that can be set on the Tab item is not used at all.

tcbegley:
Hi @mikesmith1611 

It appears to be working for me, see this minimal example

```python
import dash
import dash_bootstrap_components as dbc

app = dash.Dash(external_stylesheets=[dbc.themes.BOOTSTRAP])

app.layout = dbc.Container(
    dbc.Tabs(
        [
            dbc.Tab("This is tab 1", label="tab 1"),
            dbc.Tab("This is tab 2", label="tab 2", style={"padding": "10px"}),
        ]
    )
)

if __name__ == "__main__":
    app.run_server(debug=True)
```

The padding is correctly applied to the second tab. This is happening [here](https://github.com/facultyai/dash-bootstrap-components/blob/master/src/components/tabs/Tabs.js#L80), the `Tabs` component is responsible for rendering tabs, the `Tab` component itself is really just used like an object and when used with `Tabs` will never have it's render method called.

If you have an example where it's not working, could you share it please so we can investigate?

mikesmith1611:
Hi, my mistake. I wrongly thought that the style on the Tab component was supposed to be applied to the Li or nav-item that contains the label. I am using the Tabs in a callback so I do not set the children here so didn’t notice what was happening. It would be great if we had the ability to set a style for the nav-items. I would like them to all float left as I want to put them in a scroll bar in a situation where I have many tab items.

tcbegley:
Ah understood! Yes, having some control over the style of the tab itself sounds like a good idea, and should be pretty doable. I'll make sure this gets added, thanks for the feedback!

tcbegley:
Resolved by #152

PR #152, opened 2019-03-14
Add tab_style prop to Tab component

tcbegley:
Allows you to add inline styles to the tab (i.e. the underlying nav-link rather than just the tab content) as suggested in #151 

*EDIT* - Changed `tab_style` to `label_style` and added new `tab_style` prop to style the `NavItem` based on below feedback.

Here's a little example you can use to see this in action, works with both static tab content, and content set by callbacks.

```python
import dash

app.layout = dbc.Container(
    [
        dbc.Tabs(
            [
                dbc.Tab(
                    "This is tab 1",
                    label="tab 1",
                    label_style={"color": "#ff9900"},
                )
            ]
        ),
        dbc.Tabs(
            [
                dbc.Tab(
                    label="tab 1", label_style={"color": "#0099ff"}
                ),
                dbc.Tab(label="tab 2"),
            ],
            id="tabs",
        ),
        html.Div(id="tab-content"),
    ]
)


@app.callback(Output("tab-content", "children"), [Input("tabs", "active_tab")])
def make_tab_content(at):
    return at


if __name__ == "__main__":
    app.run_server(debug=True)
```

The tabs should get rendered something like this: 
image

@mikesmith1611 interested to hear what you think about this?

### DIFF
--- a/src/components/tabs/Tab.js
+++ b/src/components/tabs/Tab.js
@@ -19,9 +19,22 @@ Tab.propTypes = {
   children: PropTypes.node,
 
   /**
-   * Defines CSS styles which will override styles previously set.
+   * Defines CSS styles which will override styles previously set. The styles
+   * set here apply to the content of the Tab
    */
   style: PropTypes.object,
+
+  /**
+   * Defines CSS styles which will override styles previously set. The styles
+   * set here apply to the NavItem in the tab.
+   */
+  tab_style: PropTypes.object,
+
+  /**
+   * Defines CSS styles which will override styles previously set. The styles
+   * set here apply to the NavLink in the tab.
+   */
+  label_style: PropTypes.object,
 
   /**
    * Often used with CSS to style elements with common properties.
--- a/src/components/tabs/Tabs.js
+++ b/src/components/tabs/Tabs.js
@@ -56,11 +56,12 @@ class Tabs extends React.Component {
       child = child.props.children;
       const tabId = child.props.key || child.props.tab_id || 'tab-' + idx;
       return (
-        <NavItem key={tabId}>
+        <NavItem key={tabId} style={child.props.tab_style}>
           <NavLink
             className={classnames({
               active: this.state.activeTab === tabId
             })}
+            style={child.props.label_style}
             onClick={() => {
               this.toggle(tabId);
             }}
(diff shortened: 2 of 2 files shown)


mikesmith1611:
I notice that the new tab_style is set on the NavLink rather than the NavItem. Maybe we could have label_style that is applied to the NavLink and also have tab_style which is applied to the NavItem?

tcbegley:
Yes, I first tried it on the `NavLink`, but most Bootstrap styles are set on `NavItem` so it has little effect, though having said that I suppose the float example you mentioned in the other issue is exactly an example of when it might be necessary to have control other both >_<

tcbegley:
Alright, it's now possible to set inline styles of both `NavLink` and `NavItem` using `label_style` and `tab_style` respectively.

mikesmith1611:
> Yes, I first tried it on the `NavLink`, but most Bootstrap styles are set on `NavItem` so it has little effect, though having said that I suppose the float example you mentioned in the other issue is exactly an example of when it might be necessary to have control other both >_<

My thoughts exactly! Thanks that was fast. I am slowly moving over to dash-bootstrap-components from just setting bootsrap classes on div elements. It's a nice package I appreciate your efforts !

mikesmith1611:
Is this likely to be merged soon?

tcbegley:
I'll give others a chance to look over this before we merge into master, but in the meantime I made a prerelease that you can use:

```
pip install dash-bootstrap-components==0.3.6rc1
```

pbugnion:
This is great, thank you!

pbugnion:
Would it be worth opening another issue to add instructions to the documentations? Not sure how discoverable this will be otherwise?

tcbegley:
> Would it be worth opening another issue to add instructions to the documentations? Not sure how discoverable this will be otherwise?

Yes sure. They are at least in the generated api documentation, but perhaps some specific examples would be useful too.

polystethylene:
Just in the process of converting a layout written with standard Dash dcc and html components to dbc, and have come a bit unstuck replicating different styles for tabs dependent on which one is selected.

tcbegley:
Hey @polystethylene 

Currently there isn't a `selected_style` or anything equivalent, though we could certainly look at adding something like that.

For now you can style the selected tab using some custom CSS. The tabs container has the `nav-tabs` class applied, while all the tabs have the `nav-link` class applied. The active tab will additionally have the `active` class applied (these are all just the classes that the Bootstrap stylesheet itself uses, but you can override any of them). Hence, if I write some CSS like this

```css
.nav-tabs .active.nav-link {
  background:blue;
  color:white!important;
}
```

I get something like this

image

polystethylene:
Ah yes of course - great shout thanks!

Thanks for all your work on this project, I'm enjoying adding way more clarity to the structure of my dash apps :)

PR #154, opened 2019-03-16
Fix setting props on input components

pbugnion:
Prior to this commit, changing the value prop on `Input` didn't result in the value of the underlying DOM input changing. In practice, this meant callbacks that changed the `value` attribute of the Input did nothing.

As a minimal reproducible example:

```py
import dash

app = dash.Dash(__name__)

app.layout = dbc.Container(
    [
        dbc.Input(id='example-input'),
        dbc.Button(id='example-button')
    ]
)

@app.callback(
    Output('example-input', 'value'),
    [Input('example-button', 'n_clicks')]
)
def _cb(n_clicks):
    return str(n_clicks)


if __name__ == '__main__':
    app.run_server(port=8888)
```

The underlying problem is that `value` is only transferred from the props to the state on initial construction, not when the input receives new props. This fix changes the behaviour so that the state is always set.

This PR also:
- adds tests for the behaviour of the Input component when the props change or when events are fired (looking specifically at the interaction of `setProps` and `debounce`).
- improves the wording of the `debounce` prop (which I think is a poor name, but reasonable since it matches dcc's input).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "dash-bootstrap-components",
-  "version": "0.3.5",
+  "version": "0.3.6-dev",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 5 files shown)


tcbegley:
This is great, thanks for taking care of it!

PR #158, opened 2019-03-21
Dismissable alerts

tcbegley:
This PR adds a `dismissable` prop to `Alert` which, when set to `True`, adds a dismiss button to the alert. It can be used alongside the existing `is_open` prop to display or dismiss alerts. See the following screenshot:

image

I have also taken the opportunity to improve the alert documentation, and add some more detailed examples.

### DIFF
--- a/docs/components_page/components/alert/__init__.py
+++ b/docs/components_page/components/alert/__init__.py
@@ -1,19 +1,79 @@
 from pathlib import Path
 
+import dash_core_components as dcc
 import dash_html_components as html
 
 from ...api_doc import ApiDoc
-from ...helpers import ExampleContainer, HighlightedSource
+from ...helpers import (
+    ExampleContainer,
+    HighlightedSource,
+    load_source_with_environment,
+)
 from ...metadata import get_component_metadata
-from .alert import alerts
+from .content import alert as alert_content
+from .link import alerts as alerts_link
+from .simple import alerts as alerts_simple
 
 HERE = Path(__file__).parent
 
-source = (HERE / "alert.py").read_text()
+alerts_simple_source = (HERE / "simple.py").read_text()
+alerts_link_source = (HERE / "link.py").read_text()
+alert_content_source = (HERE / "content.py").read_text()
+alerts_dismiss_source = (HERE / "dismiss.py").read_text()
 
-content = [
-    html.H2("Alert"),
-    ExampleContainer(alerts),
-    HighlightedSource(source),
-    ApiDoc(get_component_metadata("src/components/Alert.js")),
-]
+
+def get_content(app):
+    return [
+        html.H2("Alerts", className="display-4"),
+        html.P(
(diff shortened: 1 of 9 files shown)


docs/components_page/components/alert/__init__.py
```diff
@@ -1,19 +1,79 @@
...
+        ),
+        ExampleContainer(alert_content),
+        HighlightedSource(alert_content_source),
+        html.H4("Dismissing"),
+        html.P(
+            dcc.Markdown(
+                "Set `dismissable=True` to add a dismiss button to the alert "
+                "which closes the alert on click. You can also use the "
+                "`is_open` property in callbacks to show or hide the alert. "
+                "By default the alert appeaers and disappears with a fade "
```
pbugnion: appears

docs/components_page/components/alert/__init__.py
```diff
@@ -1,19 +1,79 @@
...
+                "animation, to disable this simply set `fade=False`."
```
pbugnion: I feel like that sentence could benefit from being split into two:

> By default the alert appeaers and disappears with a fade animation. To disable this simply set fade=False.

pbugnion:
For reference, this is what the updated documentation looks like:

screenshot

pbugnion:
Looks great!

jchang10:
I just noticed that 'dismissable' only works if you give the alert an id. Seems like an unnecessary burden to have to give an id to an Alert that will never get used on the server. Only needed to hide it on the client, right?

Just for future reference, I wonder, is this a quirk about Dash not being able to propagate events to id-less components?

tcbegley:
Hey @jchang10,

Thanks a lot for raising this. I haven't been able to reproduce it though. This is the test app I ran:

```python
import dash

app.layout = dbc.Alert("This is an alert", dismissable=True)

if __name__ == "__main__":
    app.run_server(debug=True, port=8888)
```

The Alert dismissed as I would have expected when clicking on the dismiss button. These are the versions of Dash libraries I was using:

```
dash==0.41.0
dash-bootstrap-components==0.4.0
dash-core-components==0.46.0
dash-html-components==0.15.0
dash-renderer==0.22.0
dash-table==3.6.0
```

Could you share an example where you see the behaviour you describe?

jchang10:
Yes, exact same versions, but I am still seeing the same problem. 

The only difference with my test is this line. Change:
    app.layout = dbc.Alert("This is an alert", dismissable=True)
to:
app.layout = html.Div( dbc.Alert("This is an alert", dismissable=True), id='testdiv')

IE, add a Div wrapper or even a dbc.Card, etc. If I add an id to Alert, it works, tho.

tcbegley:
@jchang10 I managed to reproduce it, thanks. Also, that is very weird! I'll try to figure out what the problem is.

I'm going to make an issue for this for visibility.

Issue #148, opened 2019-03-08
Kill DatePickerSingle, DatePickerRange.


tcbegley:
These two components make our bundle a lot larger through the `react-dates` dependency, but ultimately just replicate functionality present in *dash-core-components*. I think we should kill them and instead encourage use of the *dash-core-components* version, in conjunction with the CSS being added in #147.

Neither component is explicitly documented (though `DatePickerSingle` does appear in the inline form example on the forms page of the docs.). Nevertheless, it would be good to decide on the best way to deprecate these.

pbugnion:
Removing these sounds good. One way to go about removing them could just be to replace them with something like:

```py
class DatePickerSingle(object):

    def __init__(self, *args, **kwargs):
        raise AttributeError("DatePickerSingle has been removed. Use DatePickerSingle from dash-core-components with the attribute className=\"dash-bootstrap\"")
```

Doing this would save an annoying deprecation cycle.

tcbegley:
Resolved by #159

Issue #60, opened 2018-11-17
Reduce the bundle size


pbugnion:
The generated bundle is fairly big. I think this is because we include parts of modules that we don't need.

I don't think this is urgent at all, but if we can, it might be worth spending some time understanding how to reduce the bundle size.

tcbegley:
I believe #148 should help with this.

pbugnion:
Definitely. I haven't tested this, but I have a hunch that we can also reduce the bundle size considerably by replacing blanket imports of *ramda* with more targeted imports.

See e.g. https://github.com/ramda/ramda#usage.

tcbegley:
#159 reduced the bundle size to about a third of what it was. At 268KB it's still slightly above the recommended limit of 244KB, but perhaps that's fine? There may be some scope to reduce a bit further.

pbugnion:
I think the current bundle is fine. I think there's still some value in reading around how to manage the bundle size, but it's probably OK for now.

Happy to close this if you want.

tcbegley:
Agreed, happy for you to close this

PR #159, opened 2019-03-21
Kill date pickers

tcbegley:
This PR removes `DatePickerSingle` and `DatePickerRange` as suggested in #148. It appears to have reduced the bundle size by about two thirds (!) from ~750KB to ~250KB.


tcbegley:
While looking into `ramda` imports as suggested [here](https://github.com/facultyai/dash-bootstrap-components/issues/60#issuecomment-472301899) I noticed that these two components were also the only places we were using `ramda` without targeted imports, so possibly a large chunk of the reduction we're getting is due to the blanket imports getting removed.

pbugnion:
To help people use `DatePicker{Single,Range}` from dash-core-components, should we write an example in the documentation somewhere? The deprecation exception could then point to that?

pbugnion:
Apart from that, this looks good.

tcbegley:
> To help people use DatePicker{Single,Range} from dash-core-components, should we write an example in the documentation somewhere? The deprecation exception could then point to that?

We could just link to the [dash-core-components docs](https://dash.plot.ly/dash-core-components/datepickersingle)? I'm not sure that our own documentation would add much?

ZeroCool2u:
@tcbegley @pbugnion Is there any documentation on how to apply the theme to the DateRangePicker components? I haven't had any luck so far.

tcbegley:
hey @ZeroCool2u, you can try this https://github.com/tcbegley/dash-bootstrap-css

ZeroCool2u:
@tcbegley Great, thank you!

PR #160, opened 2019-03-21
Use callback_context instead of *_timestamp props

tcbegley:
This small PR updates one of the `InputGroup` examples in the docs to use `dash.callback_context` instead of `*_timestamp` props, which is now the preferred approach. See [the Dash FAQs](https://dash.plot.ly/faqs)

> Prior to v0.38.0, you needed to compare timestamp properties like n_clicks_timestamp to find the most recent click. While existing uses of *_timestamp continue to work for now, this approach is deprecated, and may be removed in a future update.

### DIFF
--- a/docs/components_page/components/input_group/dropdown.py
+++ b/docs/components_page/components/input_group/dropdown.py
@@ -1,3 +1,4 @@
+import dash
 import dash_bootstrap_components as dbc
 from dash.dependencies import Input, Output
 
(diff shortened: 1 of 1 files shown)


docs/components_page/components/input_group/dropdown.py
```diff
@@ -22,21 +23,22 @@
...
     [
-        Input("dropdown-menu-item-1", "n_clicks_timestamp"),
-        Input("dropdown-menu-item-2", "n_clicks_timestamp"),
-        Input("dropdown-menu-item-clear", "n_clicks_timestamp"),
+        Input("dropdown-menu-item-1", "n_clicks"),
+        Input("dropdown-menu-item-2", "n_clicks"),
+        Input("dropdown-menu-item-clear", "n_clicks"),
     ],
 )
 def on_button_click(n_clicks_t1, n_clicks_t2, n_clicks_tclear):
```
pbugnion: Does it make sense to change the argument names? Most commonly, unused arguments are replaced by a `_`, but obviously that won't work for three buttons.

pbugnion: At least removing the `t` in the argument name since we're catching `n_clicks`, not `n_clicks_timestamp`.

pbugnion:
Looks great -- it's much more understandable. Merge when you're ready.

docs/components_page/components/input_group/dropdown.py
tcbegley: Good point, I'll amend

Issue #164, opened 2019-03-25
Tab component does not work with dash v0.40


vivekvs1:
Hi,

After updating `Dash `to `v0.40`, the Tab component results in '`Error Loading Layout`'.  I have confirmed that the Tab component works well if I switch back to `v0.39`.

Thanks,
Vivek

tcbegley:
Hi @vivekvs1 

Thanks for raising this. I have managed to reproduce the problem, and will try to understand what's changed and why things are breaking now as soon as possible. Sorry for the inconvenience.

vivekvs1:
Thank you :)

tcbegley:
Resolved by #166

vivekvs1:
Thank you!!

tcbegley:
No worries, thanks for alerting us to this in the first place!

PR #166, opened 2019-03-26
Fix Tabs in Dash 0.40.0

tcbegley:
This PR addresses #164.

Background
When using `Tabs` and `Tab`, each `Tab` is not really a child of `Tabs` in the usual sense. Instead `Tabs` iterates over the `Tab` components passed to it and extracts things like the label and content. With the release of Dash 0.40.0 the internal structure of the Dash components being iterated over changed, and the `children` prop was replaced with `_dashprivate_layout`. This PR changes the relevant references to children and updates the version requirement for Dash in the setup file (this will no longer work with older versions of Dash).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ setup(
     author_email="opensource@faculty.ai",
     url="https://github.com/facultyai/dash-bootstrap-components",
     packages=find_packages(),
-    install_requires=["dash>=0.32.1", "dash-html-components"],
+    install_requires=["dash>=0.40.0"],
     include_package_data=True,
     classifiers=[
         "License :: OSI Approved :: Apache Software License",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -32,7 +32,7 @@ def _get_long_description():
     author_email="opensource@faculty.ai",
     url="https://github.com/facultyai/dash-bootstrap-components",
     packages=find_packages(),
-    install_requires=["dash>=0.32.1", "dash-html-components"],
+    install_requires=["dash>=0.40.0"],
```
tcbegley: `dash` now pins the versions of its various dependencies, including `dash-html-components`, so there is no need for us to specify it as a separate requirement.

tcbegley:
I've been testing this a bit further and there's a few things I'm not sure of yet, going to investigate further.

Issue #168, opened 2019-03-29
Expose 'disabled' property for Input


adm78:
Dash core components `Input` component has  a `disabled` property that disables the input (element becomes greyed out and un-clickable). It would be great if the bootstrap variant also exposed this, so we can set it in the layout/attach callbacks to it.

tcbegley:
Great idea @adm78, I think that should be relatively straight forward to add. Thanks for raising.

tcbegley:
Resolved by #170

PR #170, opened 2019-03-29
Expose disabled prop of Input

tcbegley:
This PR exposes the `disabled` prop of the `Input` component as suggested in #168

### DIFF
--- a/src/components/input/Input.js
+++ b/src/components/input/Input.js
@@ -12,7 +12,7 @@ class Input extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    this.setState({value: nextProps.value})
+    this.setState({value: nextProps.value});
     if (this.props.setProps) {
       this.props = nextProps;
     }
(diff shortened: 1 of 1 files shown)


pbugnion:
For reference, this is what the disabled state looks like:
Screenshot 2019-04-07 at 08 21 05

src/components/input/Input.js
```diff
@@ -168,6 +168,11 @@ Input.propTypes = {
    */
   value: PropTypes.string,
 
+  /**
+   * If true, the input is disabled and can't be clicked on.
```
pbugnion: We're annoyingly inconsistent on using `true` and `True` in the descriptions. E.g. `in_navbar` on `DropdownMenu` has `Set this to True if the dropdown is inside a navbar. Default: False.`, but then all `disabled` props use the text you suggest. As a low-priority task, we could consistently use one or the other (I think I'm somewhat more in favour of True / False since the consumers of the documentation are Python users).

pbugnion:
Looks great!

src/components/input/Input.js
tcbegley: Good point. I agree that using `True` / `False` makes the most sense.

PR #171, opened 2019-03-30
Support loading states

tcbegley:
This PR addresses #165 by adding a `loading_state` prop to all components, which in turn is used to set the `data-dash-is-loading` attribute.

This makes *dash-bootstrap-components* compatible with the new `Loading` component in *dash-core-components*. Here's a simple example:

```python
import time

import dash

app.layout = dbc.Container(
    [
        dbc.Button("Load", id="button"),
        dbc.Card(
            [
                dbc.CardHeader("Card header"),
                dcc.Loading(dbc.CardBody(id="content")),
            ],
        ),
    ]
)


@app.callback(Output("content", "children"), [Input("button", "n_clicks")])
def load(n):
    if n:
        time.sleep(1)
        return f"The card content has been reloaded {n} times"
    return "The card content has not yet been reloaded"


if __name__ == "__main__":
```

I made a prerelease for testing this: `pip install dash-bootstrap-components==0.3.7rc1`

Happy to take ideas on how to test this properly, I've tested a lot of components so far, but not all.

### DIFF
--- a/src/components/Alert.js
+++ b/src/components/Alert.js
@@ -28,12 +28,21 @@ class Alert extends React.Component {
   }
 
   render() {
-    const {children, dismissable, is_open, ...otherProps} = this.props;
+    const {
+      children,
+      dismissable,
+      is_open,
+      loading_state,
+      ...otherProps
+    } = this.props;
     return (
       <RSAlert
         isOpen={this.state.alertOpen}
         toggle={dismissable && this.onDismiss}
         {...otherProps}
+        data-dash-is-loading={
+          (loading_state && loading_state.is_loading) || undefined
+        }
       >
         {children}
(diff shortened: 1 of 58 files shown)


pbugnion:
This looks great! Maybe in a separate PR, would it make sense to document compatibility between dash loading states and dash-bootstrap-components?

We could even make a separate component or classes that use Bootstrap's spinner?

tcbegley:
> Maybe in a separate PR, would it make sense to document compatibility between dash loading states and dash-bootstrap-components?

Agreed. It would be good to think about how we can expand the documentation past the component pages, to cover examples, custom css and things like this.

> We could even make a separate component or classes that use Bootstrap's spinner

Yes that would be great!

PR #175, opened 2019-04-04
Button / ButtonGroup documentation

tcbegley:
This PR improves the documentation for `Button` and adds separate documentation for `ButtonGroup`.

It also exposes the `group` prop of `DropdownMenu` so that it can be used inside a `ButtonGroup`, which I realised was an issue while preparing the documentation.

image

image


pbugnion:
Since we're modifying this documentation, should we also change the description of `n_clicks_timestamp` to suggest using Dash's `callback_context` instead of `n_clicks_timestamp`?

src/components/DropdownMenu.js
```diff
@@ -132,7 +132,12 @@ DropdownMenu.propTypes = {
    * Size of the dropdown. 'sm' corresponds to small, 'md' to medium
    * and 'lg' to large.
    */
-  bs_size: PropTypes.oneOf(['sm', 'md', 'lg'])
+  bs_size: PropTypes.oneOf(['sm', 'md', 'lg']),
+
+  /**
+   * Seet group to True when using inside `ButtonGroup`
```
pbugnion: Set

pbugnion: How about 

> Set group to True if the dropdown is inside a `ButtonGroup`.

?

pbugnion:
Looks great!

I made a couple of comments inline, but merge whenever.

src/components/DropdownMenu.js
tcbegley: Sounds good to me

Issue #139, opened 2019-03-03
Right-aligned menu


jmsblah:
Is it possible to get right-aligned menu on the navbar?
[Bootstrap | menu-alignment](https://getbootstrap.com/docs/4.0/components/dropdowns/#menu-alignment)

tcbegley:
Hi @jmsblah 

That's not supported currently. I think it should be possible to add a `right` prop to `dbc.DropdownMenu` and pass it to the reactstrap `DropdownMenu` [here](https://github.com/facultyai/dash-bootstrap-components/blob/master/src/components/DropdownMenu.js#L50). (Note the unfortunate clash of terminology, what Bootstrap and reactstrap call a Dropdown we call DropdownMenu because we wanted to distinguish our component from the Dropdown in *dash-core-components*). 

I tried it just now and the right alignment works, but the sizing for the menu itself was kind of messed up, so it probably isn't quite as simple as I hoped. If you want to have a go at implementing this yourself this would be a very welcome contribution! Otherwise we'll add it to the todo list and get to it eventually.

jmsblah:
Hi!
It would be great to be able to contribute, I am just not sure on how to do it. But I will give it a try!

tcbegley:
That's great that you'd like to contribute! Here's a rough guide, we should definitely add some instructions to our README though. Before I get into it, I had another think about this, and I think the sizing issues I was seeing on my experiment were a React 15/16 issue (current version of React is 16 which is what ReactStrap is built for, but *dash-renderer* uses React 15, and it sometimes causes problems, see #65). If that's true, we'll be blocked from supporting this until [Dash v1](https://github.com/plotly/dash/issues/469) is released. However, you're more than welcome to have a go at implementing this and testing against the *dash* pre-releases: *dash==1.0.0a1*, *dash-renderer==1.0.0a2* etc. Then we'll just save it up ready for when Dash v1 is released.

Anyway, here's how you would contribute if you want to give it a go:
* Fork this repository (I see you already did).
* Edit the component in `src/components/DropdownMenu.js`. As I mentioned above, I think you just need to add `right` prop to `DropdownMenu` and pass it to `RSDropdownMenu`.
* Build the components with `npm run build-dist` (first time you might have to `npm install` to get the dependencies), then install the module locally to test (`pip install -e .` from the root of the repository).
* Push your changes to your fork, then open a pull request and we'll review.

jmsblah:
[...]

PR #177, opened 2019-04-11
Updates for React 16

tcbegley:
With the release of *dash* 0.41.0 and *dash-renderer* 0.22.0, *dash-renderer* is now using React 16, meaning the work that was happening on the *dash-v1* branch can be merged.

Changes include:
- New `Modal` components + documentation (#146)
- New `Spinner` component + documentation (#161)
- Allow right-alignment of `DropdownMenu` (#139)
- Update to reactstrap 8

I made a prerelease for testing:

```
pip install dash-bootstrap-components==0.5.0rc1
```


docs/components_page/components/modal/__init__.py
```diff
@@ -0,0 +1,127 @@
...
+            load_source_with_environment(
+                modal_backdrop_source, "modal", {"app": app}
+            )
+        ),
+        HighlightedSource(modal_backdrop_source),
+        html.H4("Scrolling long content"),
+        html.P(
+            dcc.Markdown(
+                "When modals become too long for the user’s viewport or "
+                "device, they scroll independent of the page itself. If you "
```
pbugnion: independently?

docs/components_page/components/modal/__init__.py
```diff
@@ -0,0 +1,127 @@
...
+                "prefer you can specify `scrollable=True` to scroll only the "
+                "body of the model."
```
pbugnion: modal

docs/components_page/components/modal/scrollable.py
```diff
@@ -0,0 +1,55 @@
...
+from dash.dependencies import Input, Output, State
+
+modal = html.Div(
+    [
+        dbc.Button("Scrolling modal", id="open-scroll", className="mr-1"),
+        dbc.Button("Modal with scrollable body", id="open-body-scroll"),
+        dbc.Modal(
+            [
+                dbc.ModalHeader("Header"),
+                dbc.ModalBody(...),
```
pbugnion: that doesn't look right?

pbugnion: ah never mind -- this isn't actually run.

pbugnion: Out of interest, since we're just using the variable LOREM, what's the rationale for having this file separately?

tcbegley: That's a fair point. I didn't want to include the text in the example of course, but I guess injecting the variable `LOREM` into the example is no less clear than the ellipsis and means we only have one copy of the file. Is that what you had in mind?

docs/components_page/components/modal/__init__.py
pbugnion: Is it worth adding a sentence like 'By default, the entire modal (including its header) scroll'? I found that text a bit unclear.

pbugnion:
On modals, I expect people will spend a ton of time writing annoying little callbacks like:

```py
@app.callback(
    Output("modal", "is_open"),
    [Input("open", "n_clicks"), Input("close", "n_clicks")],
    [State("modal", "is_open")],
)
def toggle_modal(n1, n2, is_open):
    if n1 or n2:
        return not is_open
    return is_open
```

Could we potentially provide a function like:

```py
dbc.interaction.bind_modal_callbacks(app, 'modal')
```

?

(we could imagine similar callbacks in the `dbc.interaction` module for other cases where the interaction is obvious, like popovers and collapse).

I don't think that's in scope for this PR, but it might be worth considering.

pbugnion:
This looks great! Modulo a couple of comments inline, merge whenever!

docs/components_page/components/modal/scrollable.py
pbugnion: yes, exactly

tcbegley:
Thanks for reviewing!

I think your `dbc.interaction` module idea is nice, definitely worth experimenting with that.

PR #189, opened 2019-04-26
Input props updates

tcbegley:
This PR makes two changes:

1) Some props of `Input` that map to HTML attributes were previously not camel cased and hence not applied correctly. While we could have built some kind of shim to pass these down to the underlying HTML elements, *dash-core-components* chose to rename them, and I think it makes sense to stay consistent with them. See [their changelog](https://github.com/plotly/dash-core-components/blob/master/CHANGELOG.md#0470---2019-04-25)

2) Use timestamp rather than date object in `*_timestamp` props in `Input`.

### DIFF
--- a/src/components/input/Input.js
+++ b/src/components/input/Input.js
@@ -182,14 +182,14 @@ Input.propTypes = {
    * This attribute indicates whether the value of the control can be
    * automatically completed by the browser.
    */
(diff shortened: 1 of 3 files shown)


tcbegley:
The tests are failling because `*_timestamp` props no longer hold a date object. Will update the test since this is intended behaviour.

Issue #190, opened 2019-04-27
Add auto_dismiss function in alert


happyshows:
Hi,

I was looking for some features like shiny notification (https://shiny.rstudio.com/articles/notifications.html)

Which will dismiss msg after the duration is over. Is it possible to implement in this package?

tcbegley:
Thanks for the suggestion. I'll investigate how this is implemented in RShiny and whether we could do something similar.

It may be that [toasts](https://getbootstrap.com/docs/4.3/components/toasts/#basic) which we haven't added yet would be the best components on which to build something like this.

tcbegley:
Hey, I'm afraid not, I haven't had time to look at this. I'll try to prioritise it, but would also welcome contributions from anyone that wants to take a pass at it.

tcbegley:
This is available for both `Alert` and `Toast` in version 0.7.1 of _dash-bootstrap-components_. They each have a new `duration` prop through which you can supply a duration in milliseconds after which the `Alert` / `Toast` will dismiss. Check out the examples in the docs for `Alert` and `Toast`. 

Sorry it took a while for me to get around to it.

PR #250, opened 2019-09-06
Alert and Toast timeout

tcbegley:
This PR addresses #190, adding a `duration` prop to both `Alert` and `Toast` allowing user to specify a timeout in milliseconds after which the alert or toast is dismissed.

### DIFF
--- a/docs/components_page/components/alert/__init__.py
+++ b/docs/components_page/components/alert/__init__.py
@@ -20,6 +20,7 @@ alerts_simple_source = (HERE / "simple.py").read_text()
 alerts_link_source = (HERE / "link.py").read_text()
 alert_content_source = (HERE / "content.py").read_text()
 alerts_dismiss_source = (HERE / "dismiss.py").read_text()
+alerts_auto_dismiss_source = (HERE / "auto_dismiss.py").read_text()
 
 
 def get_content(app):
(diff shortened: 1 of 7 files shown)


tcbegley:
This can be tested by installing the pre-release

```
pip install dash-bootstrap-components==0.7.1rc3
```

Here's a simple example app for testing:

```python
import dash
from dash.dependencies import Input, Output, State

app.layout = dbc.Container(
    [
        dbc.Button("Toggle alert", id="alert-button"),
        dbc.Button("Toggle toast", id="toast-button"),
        dbc.Alert(
            "This is an alert",
            duration=3000,
            is_open=True,
            id="alert",
            dismissable=True,
        ),
        dbc.Toast(
            "This is a toast",
            header="I will dismiss",
            duration=3000,
            id="toast",
            is_open=True,
            dismissable=True,
        ),
    ]
)


def toggle(n, is_open):
    if n:
        return not is_open
    return is_open


app.callback(
    Output("alert", "is_open"),
    [Input("alert-button", "n_clicks")],
    [State("alert", "is_open")],
)(toggle)

app.callback(
    Output("toast", "is_open"),
    [Input("toast-button", "n_clicks")],
    [State("toast", "is_open")],
)(toggle)


if __name__ == "__main__":
    app.run_server(debug=True)
```

@happyshows if you have a chance to try this out and let me know what you think that would be much appreciated

Issue #173, opened 2019-04-02
Better support for `Badge` in callbacks


tcbegley:
`Badge` has a `href` prop, which when set adds hover and focus states to the badge making it actionable. However, currently it's not really possible to use the badge as an input to a Dash callback.

I propose the following two improvements to `Badge`:
- Add `n_clicks` prop to capture click events. Refer to [implementation in `Button`](https://github.com/facultyai/dash-bootstrap-components/blob/master/src/components/Button.js#L9-L16).
- Add `external_link` property and use our private `Link` component if `href` is set so that badges can be used as a Dash style link (modify the pathname without refreshing the page) for navigation in multi-page apps. Refer to [implementation in `DropdownMenuItem`](https://github.com/facultyai/dash-bootstrap-components/blob/master/src/components/DropdownMenuItem.js), as well as `ListGroupItem` and `NavLink`.

miqh:
@tcbegley, you can probably close this off now because of the work done in #268?

In hindsight, I probably should've stuck with one of the recognised keywords in the commit message to auto-close this linked issue on the merge.

PR #268, opened 2019-10-23
Improve badge for click tracking and links

miqh:
Hello,

These changes aim to address the improvements suggested in #173.

I'm aware `n_clicks_timestamp` wasn't requested, but I figure I'd include it for consistency since most other components supporting `n_clicks` also has the former.

I transplanted the documentation for the new prop types from existing component implementations.

I've also introduced (couldn't find an existing one) a test file for the badge component that covers the new behaviour introduced.


src/components/Badge.js
```diff
@@ -1,13 +1,28 @@
...
+    if (setProps) {
+      setProps({
+        n_clicks: props.n_clicks + 1,
+        n_clicks_timestamp: Date.now()
+      });
+    }
+  };
+
+  const useLink = href != null;
+  otherProps[useLink ? 'preOnClick' : 'onClick'] = incrementClicks;
```
tcbegley: The first line here is just checking if `href` is falsy, so I think it's clearer to not bother with `useLink` and have

```js
otherProps[href ? 'preOnClick' : 'onClick'] = incrementClicks;
```

This is also closer to what is done by e.g. [Button](https://github.com/facultyai/dash-bootstrap-components/blob/master/src/components/Button.js#L24), just using `href` as a truth value.

src/components/Badge.js
```diff
@@ -1,13 +1,28 @@
...
+
   return (
     <RSBadge
-      {...omit(['setProps'], otherProps)}
+      tag={useLink ? Link : undefined}
```
tcbegley: Similarly if you make the above change you could change this to

```js
tag={href && Link}
```

Issue #287, opened 2019-11-29
DropdownMenuItem as link prevents DropdownMenu dismissing


tcbegley:
If `DropdownMenuItem` is used as an internal link it prevents the `DropdownMenu` from being dismissed on click.

See [relevant thread](https://community.plot.ly/t/make-dropdown-menu-disappear-after-link-is-clicked/32021) on Plotly forum.

jezlax:
I literally came here to post this today. So wanted to +1 it, I guess. I use almost all of the dropdownMenuItems as href's across my apps.

tcbegley:
@jezlax would you mind testing out the prerelease that addresses this issue? It works at my end but it's always good to have it tested on a few different setups before releasing.

```
pip install dash-bootstrap-components==0.7.3rc1
```

Note that in addition to fixing the dismissal issue, `DropdownMenuItem` now has a property `toggle` that can be used to control whether that item does or doesn't dismiss the menu on click.

jezlax:
Thanks Tom,

Will test today and get back to you, as time permits. Thanks for the quick
fix!

jezlax:
hi Tom. I'm getting the following when i go to pip install the dev package... 

   raise BadZipFile("File is not a zip file")
zipfile.BadZipFile: File is not a zip file

it very well could be something on my end. I'll keep toying with it

tcbegley:
Hmm, that's strange... What OS / Python version are you using? Can you try in a fresh virtual environment  maybe?

jezlax:
I'll do that. windows, python 3.7.3

tcbegley:
Might also be worth trying

```
pip install --no-cache-dir dash-bootstrap-components==0.7.3rc1
```

jezlax:
Boom. Works beautifully. Looks like it deprecates me to dcc 1.0.0 on that
dbc version, so I have to revert back. Not sure if that's just an issue on
my end, but the dropdowns worked as expected in the dev environment.

tcbegley:
Great! Glad to hear it's resolved the issue for you

> Looks like it deprecates me to dcc 1.0.0 on that dbc version

That is surprising, since dcc is not a dependency of dbc, so it shouldn't require that. Can you update dcc without breaking dbc `pip install -U dash-core-components`? Keep me posted if you figure out what's going on.

jezlax:
This is what I get when i pip install the dev package...

Installing collected packages: dash-bootstrap-components,
dash-core-components
[...]

PR #288, opened 2019-12-01
Fix DropdownMenu dismissal

tcbegley:
Previously `DropdownMenuItem`s that were being used as internal links did not correctly dismiss the dropdown menu. This is because the internal links call `preventDefault` on the click event to stop the page from refreshing. However this would stop the event from reaching the underlying reactstrap component which would handle the dismissal.

This fix checks if the `DropdownMenuItem` is being used as an internal link, and if so takes care of dismissing the `DropdownMenu` itself.

To give users even more control, DropdownMenuItem now has a new prop `toggle` (`True` by default), which determines whether clicking on the item dismisses the menu.

This addresses #287


tcbegley:
This can be tested out with

```
pip install dash-bootstrap-components==0.7.3rc1
```

PR #291, opened 2019-12-20
Update dependencies

tcbegley:


tcbegley:
The force pushes were all because Git mangled the history when I accidentally clicked "update branch"

PR #299, opened 2020-01-08
Fix Input Validation

benpgreen:
This PR fixes a bug where zero values were being converted to null values in the Input component when the `type` property was set to "number".


benpgreen:
LGTM now @tcbegley

Issue #391, opened 2020-05-31
Warning: A component is changing an uncontrolled input of type radio to be controlled. (dbc.RadioButton)


np-8:
Please fill out the below information as much as possible.

- dash version: `1.12.0`
- dash-bootstrap-components version: `0.10.1`
- components affected by bug: dbc.RadioButton + dbc.Label + dbc.FormGroup

### What is happening?

Browser console shows warnings:
![image](https://user-images.githubusercontent.com/17479683/83351495-27935e00-a34d-11ea-90b4-af59eb57e604.png)

### What should be happening?

There should be no warnings in the browser console.

### Code

Copied from the example in the [docs](https://dash-bootstrap-components.opensource.faculty.ai/docs/components/input/):

```python
import dash
import dash_bootstrap_components as dbc
import dash_html_components as html
from dash.dependencies import Input, Output

app = dash.Dash(__name__)

standalone_radio_check = html.Div(
    [
        dbc.FormGroup(
            [
                dbc.Checkbox(
                    id="standalone-checkbox", className="form-check-input"
                ),
                dbc.Label(
                    "This is a checkbox",
                    html_for="standalone-checkbox",
                    className="form-check-label",
                ),
            ],
            check=True,
        ),
        dbc.FormGroup(
            [
                dbc.RadioButton(
                    id="standalone-radio", className="form-check-input"
                ),
                dbc.Label(
                    "This is a radio button",
                    html_for="standalone-radio",
                    className="form-check-label",
                ),
            ],
            check=True,
        ),
        html.Br(),
        html.P(id="standalone-radio-check-output"),
    ]
)


@app.callback(
    Output("standalone-radio-check-output", "children"),
    [
        Input("standalone-checkbox", "checked"),
        Input("standalone-radio", "checked"),
    ],
)
def on_form_change(checkbox_checked, radio_checked):
    if checkbox_checked and radio_checked:
        return "Both checked."
    elif checkbox_checked or radio_checked:
        return "One checked."
    else:
        return "None checked."



app.layout = standalone_radio_check


if __name__ == '__main__':
    app.run_server(debug=True)
```

### Error/Warning messages
**Warning 1: Uncontrolled input**
```
[...]

PR #393, opened 2020-06-06
[WIP] Functionalise

tcbegley:
This PR refactors class components to functional components, which has a couple of benefits:

- Addresses warnings generated by deprecated lifecycle methods in class components like `componentWillReceiveProps` (see #391). More generally it should be a bit more future proof as this is the favoured pattern in React now.
- Smaller bundle size!

Along the way I also:
- Improved auto-dismissal of `Alert` and `Toast`. Previously when using the `duration` prop you could set the component to auto-dismiss after a fixed period of time. However, if you dismissed manually before the time was reached, the timer would not clear, hence if the component was toggled again it would auto-dismiss early.
- Made dismissal of `DropdownMenu` when clicking on a child `DropdownMenuItem` consistent.

### DIFF
--- a/src/components/Alert.js
+++ b/src/components/Alert.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useEffect, useState} from 'react';
 import PropTypes from 'prop-types';
 import {omit} from 'ramda';
 import {Alert as RSAlert} from 'reactstrap';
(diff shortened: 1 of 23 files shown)


tcbegley:
Currently an issue here with setting values of input components like `Input`, `Select` etc. with callbacks, need to modify the approach slightly.

tcbegley:
Going to close this and try again with a cleaner commit history

PR #396, opened 2020-06-06
Function components

tcbegley:
This PR refactors class components to function components, which has a couple of benefits:

- Addresses warnings generated by deprecated lifecycle methods in class components like `componentWillReceiveProps` (see #391). More generally it should be a bit more future proof as this is the favoured pattern in React now.
- Smaller bundle size!

Along the way I also:
- Improved auto-dismissal of `Alert` and `Toast`. Previously when using the `duration` prop you could set the component to auto-dismiss after a fixed period of time. However, if you dismissed manually before the time was reached, the timer would not clear, hence if the component was toggled again it would auto-dismiss early.
- Made dismissal of `DropdownMenu` when clicking on a child `DropdownMenuItem` consistent.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,39 +5,39 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.8.3",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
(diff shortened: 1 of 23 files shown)


np-8:
Wow this looks cool! To a non-React-expert eyes looks that the code for the React components was simplified greatly.

tcbegley:
Yes! It's always been possible to write React components as functions, but until React 16.8 introduced hooks it was not possible for those components to do things like manage internal state. That's no longer true and function components are the new recommended way of doing things. I also much prefer it and agree it's a lot cleaner.

I'm going to upload a prerelease to PyPI shortly in case you want to try it out.

PR #402, opened 2020-06-12
Fixed accordion in collapse example

sarsiz:
This example doesn't work. As return statement required 3 values to be returned either in list or tuple.
So this will fix that problem.

### DIFF
--- a/docs/components_page/components/collapse/accordion.py
+++ b/docs/components_page/components/collapse/accordion.py
@@ -39,7 +39,7 @@ def toggle_accordion(n1, n2, n3, is_open1, is_open2, is_open3):
     ctx = dash.callback_context
 
     if not ctx.triggered:
-        return ['']*3
+        return False, False, False
     else:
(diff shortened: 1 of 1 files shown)


docs/components_page/components/collapse/accordion.py
```diff
@@ -39,7 +39,7 @@ def toggle_accordion(n1, n2, n3, is_open1, is_open2, is_open3):
     ctx = dash.callback_context
 
     if not ctx.triggered:
-        return ""
+        return ['']*3
```
tcbegley, line 42: I think what I originally meant to put here was `button_id = ""` rather than `return ""`, which results in none of the if statements triggering and you hit the `return False, False, False` at the end. Can you either go with that or `return False, False, False` (since `is_open` expects a boolean).

sarsiz: Yes sure

sarsiz:
Hi, 
Thank you @tcbegley for the help. I have made the changes. Please review it.

Issue #463, opened 2020-10-07
dbc.Spinner makes page reload when triggering a callback


melissadosm:
Please fill out the below information as much as possible.

- dash version: `#1.14.0`
- dash-bootstrap-components version: `#0.10.3`
- components affected by bug: callbacks

### What is happening?

I'm working with a multi-page dash app. I wrapped my main Div with a dbc.Spinner, and some of the pages on the Div have callbacks. Whenever I triggered a callback, the page reloaded the entire layout (the callback behaves correctly), and I go back to the beginning of the page. I guess what is happening is that whenever the callback triggers dash understands that it has to show the spinner.

### What should be happening?

The entire page should not reload when I trigger a callback. 

### Code

I can't provide an entire working code as it's a multipage app but I hope this gives you a sense of what was happening.

This is my main script

```
import dash
import dash_bootstrap_components as dbc
import dash_core_components as dcc
import dash_html_components as html
from dash.dependencies import Input, Output

from importlib import import_module


app = dash.Dash(
    __name__, 
    meta_tags=[{"name": "viewport", "content": "width=device-width"}], 
    title='Canopy',
    update_title='Cargando...',
    suppress_callback_exceptions=True
)

from apps import alerta_sobrecosto, alerta_transparencia, visor_datos, home_page, metodologia, concentracion, financiacion_campanias


server = app.server


# define page layout
app.layout = html.Div(
    [
        dcc.Location(id="url", refresh=False),
        # Column for user controls (SIDE BAR)
        dbc.Spinner(
            children=[
                html.Div(
                    id="content"
                ),
            ],
            size="lg", color="primary", type="border", 
            spinnerClassName='spinner',
            type=None,
            fullscreen=True,
        ),
    ]
)



# create callback for modifying page layout
@app.callback(
    Output("content", "children"),
    [Input("url", "pathname")])
def display_page(pathname):
    if pathname == "/":
        return home_page.layout
    if pathname == "/alerta-sobrecosto":
        return alerta_sobrecosto.layout
    # if not recognised, return home page
    return home_page.layout




if __name__ == "__main__":
    app.run_server(debug=True)
```

And the layouts (home_page, alerta_sobrecosto) could have any callback and the same error triggered.

tcbegley:
Hi @melissadosm 

[...]

PR #465, opened 2020-10-16
Keep children in layout when spinner is loading. Tests updated

Jason-AM:


tcbegley:
Addresses issue raised in #463

tcbegley (approved):
This looks good to me. Was there something else left to do?

Jason-AM:
I think that was all the required work. Should be good to go.

Jason-AM:
> This looks good to me. Was there something else left to do?

I think that

PR #482, opened 2020-11-26
Active NavLinks

tcbegley:
This PR enables `NavLink` to automatically detect `active` state based on the current URL. `active` can still be set to `True` or `False` to manually control whether the `NavLink` shows as active or not, or can be set to either `"exact"` or `"partial"` with the following properties

 - If `active="exact"` then the `active` styling will be applied if the current pathname exactly matches the `href` of the `NavLink`. For example, `dbc.NavLink(..., href="/my-page")` will automatically be active on `/my-page` but on no other page.
- If `active="partial"` then the `active` styling will be applied if the current pathname starts with the `href` of the `NavLink`. For example, `dbc.NavLink(..., href="/pets")` is active on `/pets`, `/pets/dogs`, `/pets/cats` etc., but not on `/random`.

### DIFF
--- a/src/components/nav/NavLink.js
+++ b/src/components/nav/NavLink.js
@@ -112,8 +112,10 @@ NavLink.propTypes = {
 
   /**
    * Apply 'active' style to this component. Set to "exact" to automatically
-   * toggle active status when pathname matches href, or to "partial" to
-   * automatically toggle on a partial match.
+   * toggle active status when the current pathname matches href, or to
+   * "partial" to automatically toggle on a partial match. Assumes that href is
+   * a relative url such as /link rather than an absolute such as
+   * https://example.com/link
    *
    * For example
    * - dbc.NavLink(..., href="/my-page", active="exact") will be active on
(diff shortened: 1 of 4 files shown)


src/components/nav/NavLink.js
```diff
@@ -88,9 +111,20 @@ NavLink.propTypes = {
...
+   * Apply 'active' style to this component. Set to "exact" to automatically
+   * toggle active status when pathname matches href, or to "partial" to
+   * automatically toggle on a partial match.
+   *
+   * For example
+   * - dbc.NavLink(..., href="/my-page", active="exact") will be active on
+   *   "/my-page" but not "/my-page/other" or "/random"
+   * - dbc.NavLink(..., href="/my-page", active="partial") will be active on
+   *   "/my-page" and "/my-page/other" but not "/random"
    */
```
Jason-AM, line 123: Is is worth having an explanation of the format an href must take? From what I can see this doesn't account for someone using `http://...`

tcbegley: Yes, this is a good call. I'll try to clarify the docstring / documentation.

PR #485, opened 2020-11-27
Python 3.9

tcbegley:


tcbegley:
Failing due to pinned dependencies specifying a version of `lxml` that is not available for Python 3.9.

This has been resolved over on the Dash repo, so easiest thing is probably to wait for the next release of Dash before merging this: plotly/dash#1457

PR #506, opened 2021-01-11
Feature/accept all colours

shwinnn:
Allows the user to pass any valid css colour to the "color" property of components, as well as the bootstrap built in colours as before.


shwinnn:
Added some tests.

tcbegley (commented):
This is looking good. A few comments:

1. Can we kill the `package-lock.json` file? The diff is unusually large (were you using Yarn?) and I normally wait until making a release to update it in any case to minimise noise in the commit history.
2. The fill colour on hover with `dbc.Button(..., outline=True)` doesn't get updated, it stays as secondary (default). This is probably hard to fix because you can't set hover styles inline.
3. Related to last point, should we perhaps steer people towards default colours in `Button`? It's not just the outline style that suffers, but also the hover effects for regular buttons get destroyed by the override here. I basically am unsure we should make it easier for people to make their app look worse 😆 

Anyway, other than that it's all looking pretty good! Here's the app i was testing with for reference. Note that I chose the colours randomly for testing haha.

```python
import dash

app.layout = html.Div(
    [
        dbc.NavbarSimple(brand="Navbar", color="red"),
        dbc.Navbar("Hello", color="blue"),
        dbc.Container(
            [
                dbc.Alert("Hello", color="chartreuse"),
                dbc.Button("Hi", color="rebeccapurple"),
                dbc.Button("There", outline=True, color="cyan"),
                dbc.Badge("Badge", color="magenta"),
                dbc.Label("Testing", color="orange"),
                dbc.Progress(
                    color="chocolate", value=50, striped=True, animated=True
                ),
                dbc.Spinner(color="darkviolet"),
                dbc.Card("This is a card", body=True, color="deepskyblue"),
                dbc.DropdownMenu(
                    [dbc.DropdownMenuItem("Item 1")],
                    label="Toggle",
                    color="lightcoral",
                ),
                dbc.FormText("Text", color="lightsteelblue"),
                dbc.ListGroup(
                    [
                        dbc.ListGroupItem("Item 1", color="fuchsia"),
                        dbc.ListGroupItem(
                            dbc.ListGroupItemText("Text", color="chocolate")
                        ),
                    ]
                ),
            ],
            className="p-5",
        ),
    ]
)

if __name__ == "__main__":
    app.run_server(debug=True)
```

shwinnn:
I've sorted out the problems with hovering at the cost of 2 new requirements. Bundle size only goes up to 302KB though. What do you think?

shwinnn:
Reverted the hover colours, button is back to only allowing bootstrap contextual colours.

src/components/__tests__/Button.test.js
```diff
@@ -40,14 +40,6 @@ describe('Button', () => {
...
   });
 
-  test('applies outline styles with "outline" prop', () => {
-    const {
-      container: {firstChild: button}
-    } = render(<Button color="primary" outline />);
-
-    expect(button).toHaveClass('btn-outline-primary');
-  });
-
```
tcbegley, line 50: Could we leave this in?

tcbegley (changes requested):
This looks great. If you could just put the Button test back in (looks like it got deleted in the back and forth on that component) I think this is ready to go.

src/components/__tests__/Button.test.js
shwinnn: For sure! I definitely accidentally deleted it.

shwinnn:
Fixed!

tcbegley (approved):
Thanks @shwinnn!

Issue #507, opened 2021-01-12
Respect multi-indices in ``dbc.Table.from_dataframe``


araichev:
It would be useful for ``dbc.Table.from_dataframe`` to a multi-index DataFrames into a table with a multi-tiered (column or row) header.
For example, format this DataFrame much like a Jupyter notebook does.

```
header = [
   np.array(['location','location','location','location2','location2','location2']), 
   np.array(['S1','S2','S3','S1','S2','S3']),
] 
df = pd.DataFrame(np.random.randn(5, 6), index=['a','b','c','d','e'], columns = header )   

df
    location                      location2                    
    S1        S2        S3         S1        S2        S3
a -1.469932 -1.544511 -1.373463  -0.317262  0.024832 -0.641000
b  0.047170 -0.339423  1.351253   0.601172 -1.607339  0.035932
c -0.257479  1.140829  0.188291  -0.242490  1.019315 -1.163429
d  0.832949  0.098170 -0.818513  -0.070383  0.557419 -0.489839
e -0.628549 -0.158419  0.366167  -2.319316 -0.474897 -0.319549
``` 

At present the ``dbc.Table.from_dataframe`` function flattens the multi-index.

tcbegley:
Nice idea, I think this should be possible.

tcbegley:
This is now supported in `0.11.2`

```
pip install -U dash-bootstrap-components
```

PR #509, opened 2021-01-13
Support multiple column levels

shwinnn:
Renders pandas MultiLevel column titles in the Table component when generated from a DataFrame.

As asked for in https://github.com/facultyai/dash-bootstrap-components/issues/507

### DIFF
--- a/dash_bootstrap_components/_table.py
+++ b/dash_bootstrap_components/_table.py
@@ -101,7 +101,7 @@ def _generate_table_from_df(
             for level in range(1, n_levels + 1)
         ]
 
-        #  Go from header break positions to cell spans
+        # Go from header break positions to cell spans
         header_spans = [
             reversed(
                 [
--- a/dash_bootstrap_components/_table.py
+++ b/dash_bootstrap_components/_table.py
@@ -78,7 +78,7 @@ def _generate_table_from_df(
         # Get the actual headers
         n_levels = df.columns.nlevels
         header_values = [
-            [val for val in df.columns.get_level_values(level)]
+            list(df.columns.get_level_values(level))
             for level in range(n_levels)
         ]
 
--- a/dash_bootstrap_components/_table.py
+++ b/dash_bootstrap_components/_table.py
@@ -1,8 +1,9 @@
 from functools import reduce
-from itertools import accumulate, groupby
+from itertools import groupby
 from operator import add
 
 import dash_html_components as html
+from numpy import cumsum
 
 
 def _generate_table_from_df(
@@ -85,7 +86,7 @@ def _generate_table_from_df(
         # The positions of header changes for each level as an integer
         header_breaks = [
             list(
-                accumulate(
+                cumsum(
                     [
                         len(list(group))
                         for _, group in groupby(header_values[level])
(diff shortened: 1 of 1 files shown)


tcbegley (commented):
It would be nice if the higher levels weren't repeated, see attached screenshot. You can achieve this using `colSpan` in the `html.Th` elements.

I had a quick look just now but I didn't see any obvious ways of extracting the span, so you might have to figure it out yourself from `df.columns.to_list()` or similar. I guess it would look something like at each level you count the number of repeated indexes, taking into account when there's a change either in the current level or the level above.

image

shwinnn:
Sorted. This will also work for column titles even in a DataFrame without multiple title 'levels' if they are repeated and adjacent. I think that's a good thing but could add logic to untangle that case if you disagree.

shwinnn:
Oh hang on - I didn't consider the case where the level *above* changes but the below level is the same. I'll think about that.

tcbegley:
Yeah, that seems to be the one outstanding thing. I've checked that if a key appears multiple times non-contiguously then it's alright.

```python
import dash
import numpy as np
import pandas as pd

header = [
    [
        "location",
        "location",
        "location",
        "location",
        "location2",
        "location2",
        "location2",
        "location3",
        "location3",
        "location",
        "location",
    ],
    ["S1", "S2", "S3", "S3", "S1", "S2", "S3", "S3", "S3", "S1", "S2"],
]
df = pd.DataFrame(np.random.randint(0, 10, size=(10, 11)), columns=header)


app.layout = dbc.Container(dbc.Table.from_dataframe(df), className="p-5",)

if __name__ == "__main__":
    app.run_server(debug=True)
```

One thing I did notice running this example is a lot of `indexing past lexsort depth may impact performance` warnings in the terminal. This can be resolved by sorting the column multiindex, but perhaps users want a particular ordering of columns which would be broken by doing this automatically

shwinnn:
I definitely don't want to reorder the rows automatically. What do you think about just suppressing the warning (or allowing just the first instance through?)

shwinnn:
Oh never mind! It can be avoided by just using `.iloc[i,j]` rather than `.loc[i][col]`

shwinnn:
So that turned out to be far from trivial to do without having abominable for loops everywhere. I tested using the same file above, but changing the second "S1" to "S3". This should force a break even though both columns in that level have the title "S3" because the level above changes from "location" to "location2" at that point.

shwinnn:
So it appears `itertools.accumulate` doesn't exist in Python 2.7. I will switch to using `numpy.cumsum`.

shwinnn:
Okay, I've just written my own accumulator instead.

tcbegley (approved):
Sorry it took me a while to look at this, looks good to me, thanks for the contribution!

Looks like there's a few conflicts with `master`. Could you merge in the changes or rebase onto `master`? Once you've done that I'll squash and merge.

shwinnn:
I didn't get any conflicts, but I've merged the changes (which seem to consist only of the colours PR I made last week).

tcbegley:
Ah yeah, there might not have been conflicts in the sense of merge conflicts, but GitHub is very conservative about letting you merge changes if there are any newer commits on the target branch. Something like it will only do a fast-forward merge, not a three-way merge? Anyway, thanks for updating!

Issue #518, opened 2021-02-12
NavLink test is flaky


tcbegley:
The NavLink tests have been intermittently failing recently. Could do with being more robust.

[Test specification](https://github.com/facultyai/dash-bootstrap-components/blob/main/tests/test_navlink.py)

PR #522, opened 2021-02-14
Make NavLink test less flaky

tcbegley:
Test was waiting for element to be present in the page, but that wasn't guaranteeing props would be set correctly by the callback. This adds an extra wait to make sure that the callback has had time to update the links.

Fixes: #518


tcbegley:
Re-ran these three times without problems so it seems more reliable than before 🤷‍♂️

Issue #526, opened 2021-02-23
Allow kwarg "download" for dbc.Button


amarvin:
`dbc.Button` is convenient for adding URI links as "href" attribute for downloadable files, where it's implemented as an `<a>` element with Bootstrap button styles. HTML5 allows `<a>` elements to have a kwarg "download" to set the desired filename for the downloaded file (https://stackoverflow.com/a/18678698/6068036), but `dbc.Button` throws an error:

```python
TypeError: The `dash_bootstrap_components.Button` component (version 0.11.1) with the ID "button-download" received an unexpected keyword argument: `download`
Allowed arguments: active, block, children, className, color, disabled, external_link, href, id, key, loading_state, n_clicks, n_clicks_timestamp, outline, size, style, target, type
```

Without the "download" kwarg set, the downloaded filename takes the default value of "download". So in my case everything is "download.xlsx".

I could use `html.A` instead with Bootstrap classes, but it'd be easier for `dbc.Button` to support this situation. What do you think?

tcbegley:
Hey @amarvin 

Added and made a pre-release. Feel free to try it out

```
pip install dash-bootstrap-components==0.11.4rc1
```

I included an example in #528, but would also be interested to check that it solves your problem if you have a chance to try it out.

amarvin:
> Hey @amarvin
> 
> Added and made a pre-release. Feel free to try it out
> 
> ```
> pip install dash-bootstrap-components==0.11.4rc1
> ```
> 
> I included an example in #528, but would also be interested to check that it solves your problem if you have a chance to try it out.

Worked perfectly, thanks!

PR #528, opened 2021-02-28
Add download prop to Button

tcbegley:
Add download prop to Button as requested in #526


tcbegley:
Here's an example. With the following directory structure

```
.
├── app.py
└── file.txt
```

and the following contents in `app.py`.

```python
# app.py
from pathlib import Path

import dash
import flask

HERE = Path(__file__).parent

app.layout = dbc.Container(
    dbc.Button(
        "Download",
        href="/file.txt",
        download="test.txt",
        external_link=True,
    ),
    className="p-5",
)


@app.server.route("/file.txt", methods=["GET"])
def send_file():
    return flask.send_from_directory(HERE, "file.txt")


if __name__ == "__main__":
    app.run_server(debug=True)
```

text file can be anything, just to test downloads.

PR #530, opened 2021-03-05
Add new  proptype to the popover: Trigger

danielgccr:
I propose the adding of a new proptype called Trigger, neccesary for closing a popover outside of its box.


tcbegley:
Thanks for the suggestion @danielgccr 

Could you share an example showing intended usage?

tcbegley:
Opened #531 in favour of this which adds support for this feature. A little more work was needed than just exposing the prop.

PR #548, opened 2021-03-23
component gallery

AnnMarieW:
Added component gallery to theme_explorer.py demo app


tcbegley (approved):
Thanks for this! I'll resolve the linting errors on my branch as I need to make a few changes to the template anyway.

PR #557, opened 2021-04-19
Added color picker example to docs

AnnMarieW:
Here is the color picker example.

### DIFF
--- a/docs/components_page/components/input/colorpicker.py
+++ b/docs/components_page/components/input/colorpicker.py
@@ -1,10 +1,10 @@
+import dash_bootstrap_components as dbc
 import dash_html_components as html
 from dash.dependencies import Input, Output
-import dash_bootstrap_components as dbc
 
 colorpicker = dbc.FormGroup(
     [
-        dbc.Label("Choose the text color:", id="label"),
+        dbc.Label(["Select a ", html.Span("color", id="color")]),
         dbc.Input(
             type="color",
             id="colorpicker",
@@ -14,9 +14,18 @@ colorpicker = dbc.FormGroup(
     ]
 )
 
-
-@app.callback(
-    Output("label", "style"), Input("colorpicker", "value"),
+app.clientside_callback(
+    """
+    function(color) {                                      
+        return {"color": color}
+    }
+    """,
+    Output("color", "style"),
+    Input("colorpicker", "value"),
 )
-def update_text(color):
-    return {"color": color}
+
+
+# Regular Dash callback
+# @app.callback(Output("color", "style"), Input("colorpicker", "value"))
+# def update_text(color):
+#     return {"color": color}
--- a/docs/components_page/components/input/colorpicker.py
+++ b/docs/components_page/components/input/colorpicker.py
@@ -1,0 +1,23 @@
+
+import dash_html_components as html
+from dash.dependencies import Input, Output
+import dash_bootstrap_components as dbc
+
+colorpicker = dbc.FormGroup(
+    [
+        dbc.Label("Choose the text color:", id='label'),
+        dbc.Input(
+            type="color",
+            id="colorpicker",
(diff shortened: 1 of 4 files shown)


docs/components_page/components/input/colorpicker.py
```diff
@@ -0,0 +1,22 @@
+import dash_html_components as html
```
tcbegley, line 1: You can delete this as it's not used in the example.

docs/components_page/components/input/colorpicker.py
```diff
@@ -0,0 +1,22 @@
...
+            id="colorpicker",
+            value="#000000",
+            style={"width": 75, "height": 50},
+        ),
+    ]
+)
+
+
+@app.callback(
+    Output("label", "style"), Input("colorpicker", "value"),
```
tcbegley, line 19: Can you remove the trailing comma after `Input`? If you run `black` on this it will explode the arguments onto multiple lines but it fits quite neatly onto 1 line here

docs/components_page/components/input/colorpicker.py
```diff
@@ -0,0 +1,22 @@
+import dash_html_components as html
+from dash.dependencies import Input, Output
+import dash_bootstrap_components as dbc
+
+colorpicker = dbc.FormGroup(
+    [
+        dbc.Label("Choose the text color:", id="label"),
```
tcbegley, line 7: If the user chooses white then the label completely disappears, maybe we could do something like

```python
dbc.Label(["Select a ", html.Span("color", id="color")])
```

and target just the word "color" with the callback?

docs/components_page/components/input/colorpicker.py
```diff
@@ -0,0 +1,22 @@
...
+        ),
+    ]
+)
+
+
+@app.callback(
+    Output("label", "style"), Input("colorpicker", "value"),
+)
+def update_text(color):
+    return {"color": color}
```
tcbegley, line 22: This is cool! But I noticed that if you drag the selector around the color picker it fires the callback continuously, which might be a little bit of a problem if lots of readers of the documentation start doing it. Could we write this one as a clientside callback so we don't spam the server too much?

I think this problem can maybe be fixed in code, so that the callback is only triggered when you let go of the selector.

tcbegley (changes requested):
Thanks @AnnMarieW! This looks like a nice addition (and thanks for adding the api docs for `Select`).

I've left a few comments inline. I also just merged #558 which should help with some of the formatting issues, you can now do

```sh
pip install nox
nox -s format
```

Finally you'll need to add `docs/components_page/components/input/colorpicker.py` [here](https://github.com/facultyai/dash-bootstrap-components/blob/dce43e768360521f65b810f6ec519f28296561c5/setup.cfg#L14-L15) as `flake8` will complain that `app` is not defined otherwise.

docs/components_page/components/input/colorpicker.py
AnnMarieW: Actually, I need to keep now it because I included your suggestion:
`dbc.Label(["Select a ", html.Span("color", id="color")])`

docs/components_page/components/input/colorpicker.py
AnnMarieW: done

docs/components_page/components/input/colorpicker.py
AnnMarieW: Good idea!

docs/components_page/components/input/colorpicker.py
AnnMarieW: Nice catch - I updated with a clienside callback, and I think it's a nice demo of this feature in Dash

tcbegley (approved):
This is great, thanks!

PR #584, opened 2021-05-19
Added R example apps and a pointer to the Github pages from the docs.

AnnMarieW:
Here are all the examples translated to R.  There is one app for each of the components.  These can be run as stand-alone apps until the snippets are ready to be included in the tabbed components and integrated into the docs.

There is a pointer on the Examples page and the Quickstart page to the R examples in Github.


tcbegley:
Thanks for this! Will be happy to merge after reviewing, just don't really have the time to do so right now. Hopefully next week I'll get around to looking a bit more carefully.

One quick request though, could you move examples to `examples/R/components` instead of creating a top-level `R_examples` folder?

AnnMarieW:
Great- no hurry!  I'll move to the folder you requested.

Issue #582, opened 2021-05-14
Adding Carousel to the components. 


MaaniBeigy:
Hi, 
I was wondering if there are plans to add bootstrap [Carousel](https://getbootstrap.com/docs/4.0/components/carousel/) component.
That would be awesome!
Thanks

tcbegley:
Sure, I don't think it would be too hard to support that. It's just something that was never prioritised.

PR #587, opened 2021-05-24
Added Carousel component. 

AnnMarieW:
This PR adds a Carousel component -- as requested in #582

This is a work in progress and I hope this draft PR will make it easier to review different API options for the component.

Currently `dcc.Carousel` is a single component.  It's possible to create separate components for the captions and headings, but so far, I find it more intuitive to format a slide with all the options in one dictionary.  

When you have some time, it might be easier to run `carousel.py`, then we can discuss in more detail.


src/index.js
```diff
@@ -66,3 +66,5 @@ export {default as Table} from './components/Table';
 export {default as Textarea} from './components/input/Textarea';
 export {default as Toast} from './components/Toast';
 export {default as Tooltip} from './components/Tooltip';
+export {default as Carousel} from './components/Carousel'
```
tcbegley, line 69: Could you move this up so they're in alphabetical order? Just makes it a bit easier to find.

src/components/Carousel.js
```diff
@@ -0,0 +1,265 @@
...
+       */
+      src: PropTypes.string,
+      /**
+       * The alternate text for an image, if the image cannot be displayed
+       */
+      alt: PropTypes.string,
+      /**
+       * The className for the image.  The default is 'd-block w-100'
+       */
+      src_className: PropTypes.string,
```
tcbegley, line 181: Can you make this `srcClassName` for consistency with other components?

src/components/Carousel.js
```diff
@@ -0,0 +1,265 @@
...
+       */
+      header: PropTypes.string,
+      /**
+       * The caption of the item.  The text is displayed in a <p> element
+       */
+      caption: PropTypes.string,
+      /**
+       * The className for the header and caption container
+       */
+      caption_className: PropTypes.string,
```
tcbegley, line 197: Similarly `captionClassName`

src/components/Carousel.js
tcbegley: And actually maybe it should be `imgClassName`?

src/components/Carousel.js
```diff
@@ -0,0 +1,265 @@
...
+       */
+      alt: PropTypes.string,
+      /**
+       * The className for the image.  The default is 'd-block w-100'
+       */
+      src_className: PropTypes.string,
+      /**
+       * The style for the image
+       */
+      src_style: PropTypes.string,
```
tcbegley, line 185: Similarly perhaps this should be `img_style`?

tcbegley (changes requested):
This looks great! Thanks so much @AnnMarieW!

I was wondering if we should add props like `n_clicks_next` and `n_clicks_previous` so that you can hook up advancing the slide to a callback? I suppose you can already connect to the `active_index` prop so maybe that's not necessary. What do you think? I'm already not so sure of the idea having written it out 😛 

Anyway, other than the minor comments I'm very happy with the component itself. It would be great if we could also add some unit tests and some documentation. For documentation most of the sections borrow heavily from the examples in the [official Bootstrap documentation](https://getbootstrap.com/docs/4.6/components/carousel/). For the tests I think checking that the children are added to the layout with the right classes, and clicking `next` etc. calls `setProps` would be great.

Happy to give pointers on either of those if needed!

AnnMarieW:
Hi @tcbegley I agree with all comments on the code - I'll make your suggested changes.

I think it's not necessary to include the `n_clicks_next` and `n_clicks_previous` props since the `active_index` can be used to control the slides in a callback.  There is an example of this in the `carousel.py` demo app.

What do you think about using the examples in the demo app for the  documentation?

tcbegley:
> I think it's not necessary to include the `n_clicks_next` and `n_clicks_previous` props since the `active_index` can be used to control the slides in a callback. There is an example of this in the `carousel.py` demo app.

Ok perfect, I agree, let's not include them!

> What do you think about using the examples in the demo app for the documentation?

Yeah they look good actually, maybe just cross-reference to Bootstrap docs to make sure we're not missing any functionality, but it looks pretty complete!

AnnMarieW:
> Yeah they look good actually, maybe just cross-reference to Bootstrap docs to make sure we're not missing any functionality, but it looks pretty complete!

One functionality that is in the Bootstrap docs that is not included is the cross-fade animation (rather than the slide animation) when changing the slides. There is no prop for that in `reactstrap` but it's possible to add just by adding `className="carousel-fade".  Should that just be added as an example?

tcbegley:
> Should that just be added as an example?

Sure why not! You could also submit a fix upstream in reactstrap if you fancy it! I did that a couple of times before when stuff wasn't supported.

I don't think the different animation styles are so important though if either of those options are a hassle.

AnnMarieW:
It would be easy to add as an example in the docs.  It looks like it will be added to  `reactstrap`  with their changes for Bootstrap 5.

docs/components_page/components/carousel.md
```diff
@@ -0,0 +1,85 @@
...
+## Crossfade
+
+The animation when changing the slide can be turned on or off using the `slide` parameter.
+Add `className="carousel-fade"` to your carousel to animate slides with a fade transition instead of a slide.
+
+{{example:components/carousel/crossfade.py:carousel}}
+
+## Control slideshow with callbacks
+
+Control which slide number is displayed by using the `active_index` in a callback. This slidshow is being controled by `dbc.RadioItems`
```
tcbegley, line 42: slidshow -> slideshow, and controled -> controlled

tcbegley (commented):
This is really shaping up!

To make the linter pass you'll need to add `docs/components_page/components/carousel/callback.py` [here](https://github.com/facultyai/dash-bootstrap-components/blob/main/setup.cfg#L8-L9)

src/components/Carousel.js
```diff
@@ -0,0 +1,250 @@
...
+    item.imgClassName =
+      typeof item.imgClassName !== 'undefined'
+        ? item.imgClassName
+        : 'd-block w-100';
+
+    return (
+      <CarouselItem
+        onExiting={() => setAnimating(true)}
+        onExited={() => setAnimating(false)}
+        key={item.id}
```
AnnMarieW, line 59: What do you think about renaming the `item` `id` prop to `key`?   I named it `id` because that's more commonly used in Dash, but people might think you can target a slide number in a callback using this `id`. (You can't)

docs/components_page/components/carousel.md
AnnMarieW: Right - thanks, I can't spell :blush:

src/components/Carousel.js
tcbegley: Yeah, if it's only being used as a `key` then I think renaming makes sense. Perhaps you can make it optional and set a default?

Something like (untested)

```js
const slides = items.map((item, idx) => {
    const key = item.key || `_dbc-carouselitem-${idx}`;
    // ...
});
```

docs/components_page/components/carousel.md
```diff
@@ -0,0 +1,72 @@
...
+  font-size: 55px;
+  color: yellow;
+}
+
+.carousel-control-prev-icon:after {
+  content: '\003c';
+  font-size: 55px;
+  color: yellow;
+}
+```
```
tcbegley, line 69: Is it worth showing the results of this with an example in the docs?

docs/components_page/components/carousel/captions.py
```diff
@@ -0,0 +1,31 @@
...
+            "header": "",
+            "caption": "This slide has a caption only",
+        },
+        {
+            "key": "4",
+            "img_style": {"height": "300px"},
+            "header": "This slide has no image",
+            "caption": "Image height must be set to make the slide visible",
+            "captionClassName": "border border-secondary text-primary",
+        },
```
tcbegley, line 29: This looks a bit strange when I run it, is that expected?

image

tcbegley (commented):
This looks great! I have two minor queries, but otherwise I think this is basically good to go 🎉

docs/components_page/components/carousel/captions.py
AnnMarieW: Yeah, this does look a little strange.  I'm not sure if it's even necessary to show how to do a slide without an image - that's not included in the official Bootstrap docs for Carousel.  (But I figured out how to do it, so I thought it would be good to include as an example)

For this use-case, it would probably be better to save any text you want displayed as an image, then include it as a slide.  Should I just delete this 4th slide?

docs/components_page/components/carousel.md
AnnMarieW: Yes, it would be good to  show the result - I just didn't know where to put this custom css. Is there an assets folder for the docs?

tcbegley: Yeah, you can add here: https://github.com/facultyai/dash-bootstrap-components/blob/main/docs/static/docs.css#L154

docs/components_page/components/carousel/captions.py
tcbegley: Yeah, I think just delete it. Not sure it would be a particularly common use case

AnnMarieW:
For now, I also removed the example of formatting the prev and next controls with CSS, since that's not part of the official Bootstrap docs either.  (Maybe because people don't do that very often?)

If someone wants to change those, I think it would be easier to hide to Bootstrap prev/next controls, replace them with a `dbc.Button` and control the slideshow with a callback.

docs/components_page/components/carousel.md
tcbegley: Did you remove this instead? Don't worry about it, we can add back in later if it's a useful example 🙂

tcbegley (approved):
Thanks @AnnMarieW! This is great 🚀

PR #596, opened 2021-06-12
R example callback tests

AnnMarieW:
Hey @tcbegley,

Here is a first attempt at creating tests for the callbacks for the R examples.   It's just a proof of concept to show how dash testing can be used to run and test the R code.   This PR is a sample test -- for the dismissable Alerts.  
Run `pytest -k test_alert_callback`

Next steps would be determine which snippets have callbacks, write the appropriate tests, use your wrapper function and read from a file rather than include the R script in the test.  

I'm looking forward to your thoughts on how to proceed.

### DIFF
--- a/docs/components_page/components/__tests__/test_snippets_callbacks.py
+++ b/docs/components_page/components/__tests__/test_snippets_callbacks.py
@@ -77,11 +77,13 @@ def test_alert_callback(dashr):
     assert dashr.find_element("#alert-fade") != []
     dashr.find_element("#alert-toggle-fade").click()
     wait.until(
-        lambda: dashr.find_elements("#alert-fade") == [], timeout=1,
+        lambda: dashr.find_elements("#alert-fade") == [],
+        timeout=1,
     )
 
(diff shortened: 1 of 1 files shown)


tcbegley:
Cool! Thanks for submitting this. I like the idea of writing a simple test for each snippet that has a callback. 

As you say I think the most important thing is that we read the snippets from the files directly. I think you can probably take some cues on how to do that from my [snippet tests](https://github.com/facultyai/dash-bootstrap-components/blob/02cbf12b0552cc1e614c9597ec09ad07ffbc2774/docs/components_page/components/__tests__/test_snippets.py) and I think they'll get automatically picked up by the CI I'm setting up.

Currently that's in a bit of a mess because I haven't figured out the right way to install dependencies for DashR testing, but once I resolve it you can rebase on top of that?

tcbegley:
It's worth rebasing on / merging #597 as I've resolved the issues with not having R properly installed in the Actions runners. Snippet tests seem to be working over there now 👌

Issue #563, opened 2021-04-30
[Docs] Tabs for R and Julia on Components page


jackparmer:
Just like the [Quickstart page](https://dash-bootstrap-components.opensource.faculty.ai/docs/quickstart/), it would be wonderful if the [Components page](https://dash-bootstrap-components.opensource.faculty.ai/docs/components/alert/) also had these tabs for R and Julia:

<img width="731" alt="image" src="https://user-images.githubusercontent.com/1865834/116744667-77403680-a9af-11eb-803b-738410c77071.png">

tcbegley:
Thanks for the suggestion @jackparmer, the plan is to add that.

Other than it being a non-trivial amount of work to translate all the examples, the main reasons this hasn't happened yet are:
- We need to update our templates to support the tabbed view. Should be easy enough but I'd ideally like to do it in a backward compatible way to allow incremental updates to the docs.
- All of those Python snippets are the actual code being run in the docs app to produce the examples, which means you have some kind of guarantee of correctness. I'd like to see if there's a way we can test all of the R and Julia examples too once added (which likely would be done in CI/CD rather than mashing together Dash, DashR and Dash.jl apps I guess)

tcbegley:
This is now live, thanks in no small part to monster translation efforts from @ann-marie-ward and @glsdown.

We're also comparing the layouts to Python in CI so R and Julia snippets are guaranteed to produce the same output as the corresponding Python snippets that are actually running in the documentation app 👌

PR #597, opened 2021-06-12
Fix R snippet tests

tcbegley:
Tests were not running properly previously. Resolved issue and fixed failing tests.


tcbegley:
This means we can now deploy R snippets to the documentation, partially resolving #563

PR #604, opened 2021-06-16
Fix docstrings, reformat reference docs

AnnMarieW:
This PR fixes how the docstrings are generated so they include the nested props such as a list of dicts.  It also updates how dictionaries and other nested props are displayed in the reference section of the docs.

Here is the before and after for `Select`.  Notice that the dictionary keys for `options` are now defined.  

**Before**
image

----
----
**After**
image

---
---
Also dictionary keys are displayed in the docs with bullet points to make them easier to read.  Here is a before and after of the  `loading_state` prop

**Before**
image

**After**
image


docs/components_page/api_doc.py
```diff
@@ -1,45 +1,49 @@
...
+    return_div = [
+        dcc.Markdown("### Keyword arguments for {}".format(component_name))
     ]
 
+    docs = component_doc.split("Keyword arguments:")[-1]
+
+    # formats code blocks that includes square brackets
+    docs = docs.replace("[", "\[").replace("]", "\]")
+    verbatim_regex = r"`((\\\[)((.|\n)*?)(\\\]))`"
+    docs = re.sub(re.compile(verbatim_regex), r"`[\3]`", docs)
```
tcbegley, line 21: I don't think efficiency is really an issue here, but each of these `compile` statements could be moved out of the function so that they only run once, not each invocation of the function. E.g.

```python
VERBATIM_PATTERN = re.compile(r"`((\\\[)((.|\n)*?)(\\\]))`")

def ComponentReference(component_name):
    ...
    docs = VERBATIM_PATTERN.sub(r"`[\3]`", docs)
```

tcbegley:
This looks so much better!

I think you will have to commit the updated `package-lock.json` because I'm using `npm ci` in the continuous integration tests which installs exact versions specified in the lock file, but that clashes with what you have in `package.json`. My bad for the incorrect advice on that 😖

docs/components_page/api_doc.py
AnnMarieW: Good idea - I moved `api_doc.py` into `markdown_parser.py` to make that easier to do.

tcbegley (approved):
Thanks!

PR #606, opened 2021-06-19
New Julia Examples

glsdown:
Added 2 new Julia examples:

- A very simple histogram based on the `faithful` python example
- The `telephones-by-region` example


glsdown:
Closing as git history was messy.

PR #607, opened 2021-06-19
Julia app examples

glsdown:
Added 3 new Julia examples:

- A very simple histogram based on the `faithful` python example
- The `telephones-by-region` example
- An adapted `wordcloud` example


examples/README.md
```diff
@@ -13,3 +13,40 @@ that can be used as a starting point for new apps.
...
+pkg> instantiate
+```
+
+Once the requirements are installed, run the file from Julia using:
+
+```julia
+julia> include("julia_file_name.jl")
+```
+
+Or you can run from any terminal using:
```
tcbegley, line 48: I think maybe any -> the since it needs to be run from the directory containing the `Project.toml` etc?

.gitignore
```diff
@@ -39,3 +39,5 @@ NAMESPACE
 deps/
 src/*.jl
 Project.toml
+!examples/**/Project.toml
+Manifest.toml
```
tcbegley, line 43: Can you add a blank line to the end of the file here?

tcbegley (changes requested):
Looks great! Could you move the simple sidebar example to `examples/julia/templates` like the current Python structure?

tcbegley (approved):

PR #620, opened 2021-07-03
Prep for move from camelCase to snake_case

glsdown:
I've been through the remaining folders in `src/components` and updated the class names from camel case to snake case. Where HTML attributes have been passed through (e.g. textarea, input), these have been updated to the html format (lowercase) instead. 

I had issues with my formatter not picking the right settings up correctly, so there may be some inconsistencies with formatting.


tcbegley (approved):
Thanks!

PR #643, opened 2021-08-14
New components - Offcanvas and pagination

glsdown:
I've added an offcanvas component (along with the relevant parts), as well as a pagination one.

There are a few items marked as todo, and a couple of fixmes in there as well for review. 

There is a known issue with the offcanvas-backdrop which means that for Bootstrap 5.1.0, the backdrop is in front of the offcanvas. Using 5.0.2 as a workaround works though.

I've added tests and documentation (hopefully), but remains untested currently as auto-tests are failing on the class_name attribute for html elements.


src/components/offcanvas/Offcanvas.js
```diff
@@ -0,0 +1,116 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {default as RBOffcanvas} from 'react-bootstrap/Offcanvas';
+
+/**
+ * Create a toggleable hidden sidebar using the Modal component.
+ * Toggle the visibility with the `is_open` prop.
```
tcbegley, line 7: This needs to be updated

src/components/offcanvas/Offcanvas.js
```diff
@@ -0,0 +1,116 @@
...
+  // - onEnter
+  // - onEntered
+  // - onEntering
+  // - onEscapeKeyDown
+  // - onExit
+  // - onExited
+  // - onExiting
+  // - onShow
+  // - restoreFocus
+  // - restoreFocusOptions
```
tcbegley, line 40: Let's drop this comment before merging.

src/components/offcanvas/Offcanvas.js
```diff
@@ -0,0 +1,116 @@
...
+   */
+  class_name: PropTypes.string,
+
+  /**
+   * The ARIA labelledby attribute
+   */
+  labelledby: PropTypes.string,
+
+  /**
+   * Includes a modal-backdrop element.
```
tcbegley, line 86: Needs to be updated to something specific to offcanvas

src/components/offcanvas/OffcanvasHeader.js
```diff
@@ -0,0 +1,56 @@
...
+ */
+const OffcanvasHeader = props => {
+  const {title, class_name, close_button, ...otherProps} = props;
+  return (
+    <RBOffcanvasHeader
+      className={class_name}
+      closeButton={close_button}
+      {...omit(['setProps'], otherProps)}
+    >
+      <RBOffcanvasTitle>{title}</RBOffcanvasTitle>
```
tcbegley, line 18: Are there any circumstances under which you would want to add something to the header that isn't a title?

I guess it's also taking control of the close button etc.

src/components/offcanvas/OffcanvasHeader.js
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Add a header to any offcanvas.
+ */
+const OffcanvasHeader = props => {
+  const {title, class_name, close_button, ...otherProps} = props;
+  return (
+    <RBOffcanvasHeader
+      className={class_name}
+      closeButton={close_button}
```
tcbegley, line 15: It would be good if the close button could be used to trigger callbacks. Perhaps the header should be incorporated into the Offcanvas component itself so that close events can be registered there?

src/components/offcanvas/__tests__/Offcanvas.test.js
```diff
@@ -0,0 +1,77 @@
...
+    for (let i = 0; i < placements.length; i++) {
+      rerender(<Offcanvas is_open placement={placements[i]} />);
+      expect(document.body.querySelector('.offcanvas-dialog')).toHaveClass(
+        `offcanvas-${placements[i]}`
+      );
+    }
+
+    // scrollable content
+    const {rerender} = render(<Offcanvas is_open />);
+    // TODO: Check body style="overflow: hidden;" when scrollable=false (default)
```
tcbegley, line 50: you can do `.toHaveStyle({overflow: "hidden"})` for this

src/components/offcanvas/Offcanvas.js
glsdown: Left here as we need to discuss which of these we might want to use.

src/components/offcanvas/Offcanvas.js
glsdown: Due to the issue with react-bootstrap, it is a `modal-backdrop` currently - this description was taken directly from their docs. However, updated to say `offcanvas-backdrop` as this is what it's called in bootstrap.

src/components/offcanvas/__tests__/Offcanvas.test.js
glsdown: Would it need to be

```js
expect(document.body.querySelector('.body')).toHaveStyle({
      overflow: 'hidden'
    });
```

or is the `.querySelector` not needed?

```js
expect(document.body).toHaveStyle({overflow: 'hidden'});
```

src/components/offcanvas/OffcanvasHeader.js
glsdown: Header I believe should only contain the title and the close button. As per the other note, I have combined Header and Body into a single component, so these behaviours are now controlled from there.

src/components/offcanvas/OffcanvasHeader.js
glsdown: There are multiple ways of dismissing the offcanvas (`is_open` property, escape key, clicking backdrop and the close button itself). In the main `Offcanvas` component, we capture these together so when it is dismissed, the `is_open` property changes. Would it be sufficient to use this to trigger callbacks, or should there be additional variations to capture each possible way of dismissing it?

Worth also mentioning that the Offcanvas component in react-bootstrap has a number of events itself, so we may potentially want to take advantage of some of these. I included them in a comment in the code, but putting it here to keep things together (and allow me to remove from the code itself):

- enforceFocus
- onEnter
- onEntered
- onEntering
- onEscapeKeyDown
- onExit
- onExited
- onExiting
- onShow
- restoreFocus
- restoreFocusOptions

In addition, it has a `container` property, which I haven't implemented.

src/components/offcanvas/__tests__/Offcanvas.test.js
glsdown: Frustratingly due to the way the tests work, it's not applying the style correctly to the body, so I'm unable to run this test. I've removed it and if required can be revisited at a later date.

tcbegley (approved):
Thanks!

PR #648, opened 2021-08-16
Update Card, Nav and Navbar for V5

AnnMarieW:
Removed CardDeck
Removed CardColumns per [react-bootstrap issue#5944](https://github.com/react-bootstrap/react-bootstrap/issues/5944)

Updates for Nav and Navbar.  
Adds new NavbarCollapse component.  The react-bootstrap Collapse doesn't has a `navbar` prop


src/components/Collapse.js
```diff
@@ -68,21 +68,11 @@ Collapse.propTypes = {
...
   /**
    * Whether collapse is currently open.
    */
   is_open: PropTypes.bool,
 
-  /**
-   * Set to True when using a collapse inside a navbar.
-   */
-  navbar: PropTypes.bool,
-
```
tcbegley, line 85: Rather than removing this and creating the new `NavbarCollapse` component, could we move the logic for that component into this file?

I think it might be possible to do something like

```jsx
const Component = navbar ? Navbar.Collapse : Collapse;

return <Component in={is_open} {...otherProps} />
```

then we can somewhat reduce the number of changes in the api. What do you think?

src/components/nav/Navbar.js
```diff
@@ -92,16 +97,18 @@ Navbar.propTypes = {
...
+   * A convenience prop for the sticky-top positioning class. Not supported in <= IE11
+   * and other older browsers
    *
    * With `sticky`, the navbar remains in the viewport when you scroll. By
    * contrast, with `fixed`, the navbar will remain at the top or bottom of
-   * the page.
+   * the page.  sticky='top'
    */
-  sticky: PropTypes.string,
+  sticky: PropTypes.oneOf(['top']),
```
tcbegley, line 111: is `'bottom'` not also an option?

src/components/nav/NavbarSimple.js
```diff
@@ -1,7 +1,8 @@
 import React, {useState} from 'react';
 import PropTypes from 'prop-types';
 import {omit} from 'ramda';
-import {Collapse, Container, Navbar as RSNavbar} from 'reactstrap';
+import {default as RBNavbar} from 'react-bootstrap/Navbar'
+import {Container} from 'react-bootstrap'
```
tcbegley, line 5: I think scoping the imports like `import Container from 'react-bootstrap'` allows for some optimisation of bundle size? I'm planning to go through everything when we're done and check all of these.

tcbegley (changes requested):
Looks great, thanks @AnnMarieW!

Left a couple of inline comments, could you rebase on top of `react-bootstrap` once you've addressed then let's get this merged!

src/components/nav/Navbar.js
AnnMarieW: No, it's actually not an option in either [Bootstrap 4](https://getbootstrap.com/docs/4.0/components/navbar/#placement) or [Bootstrap 5](https://getbootstrap.com/docs/5.0/components/navbar/#placement).

I tried it in an app and if you use `sticky="bottom"` the class shows up in the DOM but it has no effect

src/components/Collapse.js
AnnMarieW: good idea - I'll do that and remove the NavbarCollapse component

src/components/nav/Navbar.js
tcbegley: Got it, thanks! Seems strange that it's not just a boolean then 😛 

Let's leave it like this though to minimise breaking changes.

src/components/nav/NavbarSimple.js
tcbegley: ^ note the unedited version of this comment made no sense 😅

AnnMarieW: Oh, good, I was going to ask you about that...  I'll add the `/Container` - thanks!

src/components/nav/Navbar.js
AnnMarieW: So you mean it like `sticky: PropTypes.string,`   (and update the documentation so it doesn't mention bottom)
Right a boolean would be better.

tcbegley: Yeah leave it as a string since that's what it was before. Perhaps the maintainers of reactstrap and react-bootstrap made that choice in order to allow for `"bottom"` in future?

tcbegley (approved):
Amazing, thanks @AnnMarieW!

Issue #651, opened 2021-08-22
V1 documentation


AnnMarieW:
It's awesome to see the progress on the Bootstrap 5 support - there are so many cool new features. :confetti_ball: 

Here is a list of suggested changes and additions to the documentation.  Perhaps we can use this to track all pre-release issues in the docs.   This looks like a long list, but most of the items will be easy to fix.  I'm happy to get started on a PR for any of the changes you approve.  

**Home page**
- [x] On the homepage the Bootstrap example formatting is lost with the change to the Navbar. Unfortunately simply adding the container as Bootstrap 5 requires won't work as the content of the navbar is then too long. (Fixed in #676 )

**Themes Page**
- [x] Update references from Bootstrap v4 to v5 (Fixed in #686)

**Left Nav menu**
- [x] "Dash for R" just goes to the Quickstart page.  Delete? (Fixed in v1)
- [x] Add a section for how to add Bootstrap Icons (Fixed in #661 )

**Quickstart**
- [x] :arrow_up_small: Language tabs for all examples aren't working (Fixed in #654 )

**Overall**
- [x] Headings - The first subheadings are currently inconsistent. For example, sometimes it's "Basic Example" or "Basic Usage" or  "Simple Example" (or none). I think it would be better to make them all "Examples" to match the official Bootstrap documentation. (Fixed in v1)
- [x] Change all the examples back to camelCase (Fixed in #684)
- [x] Now that Dash isn't using class_name in their components, we ought to change the html ones back to className. This does depend on the note in the original bit though about delaying moving all components back to className, and how quickly this is pushed out over. (Fixed in #684)
- [x]  Make sure new components have both `className` and `class_name`. (Fixed in v1)
- [x] Add R/Julia formatters and linting (low priority) (Fixed in v1)
- [x] Links now all have text-decoration by default - needs removing
- [x] ~Question: Should we also provide some other options for styling links? Perhaps show how to add custom css to revert all links back to Bootstrap 4 behaviour? In V4, the link was underlined on hover, and in V5 it is not. I think just  removing the underline makes it less clear that it's a link. Also in V5: Removed hover and focus styles for `<a>` and `<button>` elements. We are still showing a hover effect in  `Badge`. Should this be changed in the component?~ Not changing this now. Will reconsider if people start complaining.
- [x] ~Remove lists from callback `Output, Input State` in snippets~ Worth keeping as is, as it maintains backwards compatibility, and doesn't provide that many advantages over the current approach

[...]

PR #658, opened 2021-08-28
Radio button doc example

glsdown:
As per #651 in v1 docs, there were issues with styling of the radio example in ButtonGroup docs. I have fixed this using the new class name for this purpose in the Bootstrap 5 docs. This simplified the CSS required.

When fixing it, I noticed a similar issue with the custom checked formatting example in input, so I have added some new props to the checklist and radioitems components to allow you to easily change the checked styles, without having to resort to custom CSS.

### DIFF
--- a/docs/components_page/components/button_group/radios.R
+++ b/docs/components_page/components/button_group/radios.R
@@ -6,8 +6,9 @@ button_group <- htmlDiv(
     dbcRadioItems(
       id = "radios",
       class_name = "btn-group",
-      label_class_name = "btn btn-secondary",
-      label_checked_class_name = "active",
+      input_class_name="btn-check",
+      label_class_name = "btn btn-outline-secondary",
+      label_checked_class_name = "active btn-secondary",
       options = list(
         list(label = "Option 1", value = 1),
         list(label = "Option 2", value = 2),
--- a/docs/components_page/components/button_group/radios.R
+++ b/docs/components_page/components/button_group/radios.R
@@ -8,7 +8,7 @@ button_group <- htmlDiv(
       class_name = "btn-group",
       input_class_name="btn-check",
       label_class_name = "btn btn-outline-primary",
-      label_checked_class_name = "active btn-primary",
+      label_checked_class_name = "active",
       options = list(
         list(label = "Option 1", value = 1),
         list(label = "Option 2", value = 2),
--- a/docs/components_page/components/button_group/radios.R
+++ b/docs/components_page/components/button_group/radios.R
@@ -7,8 +7,8 @@ button_group <- htmlDiv(
       id = "radios",
       class_name = "btn-group",
       input_class_name="btn-check",
-      label_class_name = "btn btn-outline-secondary",
-      label_checked_class_name = "active btn-secondary",
+      label_class_name = "btn btn-outline-primary",
+      label_checked_class_name = "active btn-primary",
       options = list(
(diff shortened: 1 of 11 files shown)


docs/static/docs.css
```diff
@@ -31,6 +31,10 @@ span.hljs-meta {
   color: #66d9ef;
 }
 
+a {
+  text-decoration: none;
+}
+
```
tcbegley, line 37: This override is very broad, if we need to remove the underline e.g. for the side navigation then I would prefer to do something more targeted / specific. It's also not clear to me that it's related to the main changes being made in this PR?

tcbegley:
```suggestion
```

docs/components_page/components/button_group/radios.R
```diff
@@ -6,8 +6,9 @@ button_group <- htmlDiv(
     dbcRadioItems(
       id = "radios",
       class_name = "btn-group",
-      label_class_name = "btn btn-secondary",
-      label_checked_class_name = "active",
+      input_class_name="btn-check",
+      label_class_name = "btn btn-outline-primary",
+      label_checked_class_name = "active btn-primary",
```
tcbegley, line 11: `btn-primary` isn't needed here, just `active`

```suggestion
      label_checked_class_name = "active",
```

docs/components_page/components/button_group/radios.jl
```diff
@@ -5,8 +5,9 @@ button_group = html_div(
         dbc_radioitems(
             id = "radios",
             class_name = "btn-group",
-            label_class_name = "btn btn-secondary",
-            label_checked_class_name = "active",
+            input_class_name = "btn-check",
+            label_class_name = "btn btn-outline-primary",
+            label_checked_class_name = "active btn-primary",
```
tcbegley, line 10: ditto

```suggestion
            label_checked_class_name = "active",
```

docs/components_page/components/button_group/radios.py
```diff
@@ -7,8 +7,9 @@
         dbc.RadioItems(
             id="radios",
             class_name="btn-group",
-            label_class_name="btn btn-secondary",
-            label_checked_class_name="active",
+            input_class_name="btn-check",
+            label_class_name="btn btn-outline-primary",
+            label_checked_class_name="active btn-primary",
```
tcbegley, line 12: ditto

```suggestion
            label_checked_class_name="active",
```

tcbegley (changes requested):
Thanks! Just a few minor comments inline.

docs/static/docs.css
glsdown: This was an accidental addition - I'll remove it.

AnnMarieW: Just FYI - this is a broad change in Bootstrap 5:
> Links are now underlined by default (not just on hover), unless they’re part of specific components.

Do we want to remove the underline by default in some  components so it doesn't have to be removed manually like in the [NavBar snippet](https://github.com/facultyai/dash-bootstrap-components/blob/v1/docs/components_page/components/navbar/navbar.py#L35) ?

glsdown: It’s such an annoying change…

I think @tcbegley broadly wants the underline removed, but didn’t like how broad this approach was as it could case the snippets to behave differently in the docs, so instead to remove on a case by case basis. Although I would definitely argue that things like the NavBar should have all links without the underline by default?

PR #662, opened 2021-08-30
Dropdown bug fixing

glsdown:
A number of bugs were identified in #651 about the dropdown. I have fixed a number of them where the props changed in the move to react-bootstrap and these are now working. However, there is a fundamental issue with the component I am unable to solve which means that the menu doesn't appear until the mouse is scrolled. Once this has been fixed, then the other issues, including those relating to the nav drop downs, should also be fixed.


docs/components_page/components/dropdown.md
```diff
@@ -47,6 +53,12 @@ Use the `active` and `disabled` arguments of `DropdownMenuItem` to set the activ
 
 {{example:components/dropdown/content.py:dropdown}}
 
+## AutoClose
```
tcbegley, line 56: This property doesn't appear to be working.

Fixing it will likely introduce a breaking change because currently dismissal of the dropdown menu is controlled by the `toggle` prop of each `DropdownMenuItem`.

I think the `auto_close` interface is nicer, but not backwards compatible so... 🤷‍♂️

tcbegley (approved):

PR #663, opened 2021-08-30
Fixed interval progress issue

glsdown:
Fixed issue with the docs example for progress component when using interval as documented in #651


tcbegley (approved):
Thanks!

PR #664, opened 2021-08-30
Changed order to maximum of 5

glsdown:
Per #651 maximum order in layout is now 5 and not 12. Updated docs to reflect this.


tcbegley (approved):
Thanks!

PR #666, opened 2021-08-31
Stop loading_state being passed to components

glsdown:
As documented in #651 , the modal components were being passed loading_state. I also identified a few other components where this was also the case, so I have fixed these.


tcbegley (approved):
Awesome, thanks!

PR #669, opened 2021-08-31
Example with collapsible sidebar and navbar

harisbal:
Thank you very much for this awesome library.
I would like to suggest including an example (for multi-page apps) where both a collapsible sidebar and a navbar are present in the layout. What I have in mind is something similar to what we see in many bootstrap templates nowadays such as [this](https://adminlte.io/themes/v3/).

In this PR I have created the barebones of the example but since I'm very weak at the css part I cannot make it look right.
image

Any help to finish this example would be much appreciated

### DIFF
--- a/examples/python/templates/multi-page-apps/responsive-collapsible-sidebar-navbar/app.py
+++ b/examples/python/templates/multi-page-apps/responsive-collapsible-sidebar-navbar/app.py
@@ -1,0 +1,171 @@
+"""
+This app creates a collapsible, responsive sidebar layout with
+dash-bootstrap-components and some custom css with media queries.
+
+When the screen is small, the sidebar moved to the top of the page, and the
+links get hidden in a collapse element. We use a callback to toggle the
+collapse when on a small screen, and the custom CSS to hide the toggle, and
+force the collapse to stay open when the screen is large.
+
+dcc.Location is used to track the current location, a callback uses the current
+location to render the appropriate page content. The active prop of each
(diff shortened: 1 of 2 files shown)


harisbal:
This is great news. Of course we should wait for the new functionality in Bootstrap 5.
Cheers!

harisbal:
I'm trying to have a go with Offcanvas but I I'm not sure it will help us achieve what I have in mind since Offcanvas disappears completely from the view.
I can't figure out how to place a collapsible sidebar (with icons and a toggle button) alongside a navbar like
this

Happy to work on this if you need help but some help with the css would be much appreciated

harisbal:
It is close but not exactly what I'm looking for. 
The main issues are two: 

1. Use of a toggle button instead of `onhover`
2. Resize the content of the page depending on the state of the sidebar (collapsed or not)

I understand that the solution is to mix the [responsive-collapsible-sidebar](https://github.com/facultyai/dash-bootstrap-components/tree/main/examples/python/templates/multi-page-apps/responsive-collapsible-sidebar) and [collapsible-sidebar-with-icons](https://github.com/facultyai/dash-bootstrap-components/tree/main/examples/python/templates/multi-page-apps/collapsible-sidebar-with-icons) but I haven't managed to make this work.

tlallema:
hello, I would also be interested to do the same things as @harisbal 
Did you make any progress on this ?
Do you have any example ?
Thank you

PR #687, opened 2021-09-10
Improved component landing page

tcbegley:
Added better example to the component landing page, and more detail on how to use it in a simple application.

### DIFF
--- a/docs/components_page/components/index/simple.jl
+++ b/docs/components_page/components/index/simple.jl
@@ -4,7 +4,8 @@ using Dash, DashBootstrapComponents
 # 2. Create a Dash app instance
 app = dash(external_stylesheets = [dbc_themes.BOOTSTRAP])
 
-# 3. Add the example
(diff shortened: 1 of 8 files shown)


docs/components_page/components/index/simple.jl
```diff
@@ -1,7 +1,20 @@
...
+badge = dbc_button(
+    [
+        "Notifications",
+        dbc_badge("4", color = "light", text_color = "primary", className = "ms-1"),
+    ],
+    color = "primary",
 )
+
+# 4. Assign your layout to the app layout
+app.layout = badge
```
AnnMarieW, line 17: How about adding a `Container` to the layout for the badge?  And maybe `fluid=True` 
Based on the question on the forum, I think  people new to Bootstrap often miss this step.

tcbegley: Good suggestion. Done!

PR #693, opened 2021-09-12
Fixed xxl column prop and added quartz to the theme list

AnnMarieW:
- Fixed a type that prevented the `xxl` prop to show in the `Col` component and updated the docs to include the new xxl size.  
- Added the quartz theme to the list of available themes in the docs.


tcbegley (approved):
Good spots, thanks!

PR #698, opened 2021-09-16
Allow dict ids as target id for Tooltip

AnnMarieW:
This PR allows the Tooltip target to include IDs that are dictionaries - like those used in pattern matching callbacks.

### DIFF
--- a/docs/content/docs/faq.md
+++ b/docs/content/docs/faq.md
@@ -6,58 +6,6 @@ title: FAQ
 
 This page contains various tips and tricks and answers to frequently asked questions about _dash-bootstrap-components_. If you think something is missing, please submit an [issue][issue] on the GitHub issue tracker.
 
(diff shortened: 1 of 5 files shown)


tcbegley (changes requested):
Nice! So great that this is working now. I swear I tried something like this with the old reactstrap implementation and it didn't like it... Anyway, not going to complain 😅 

Can you:

1. Move `stringifyId` into `Overlay.js`, then we save ourselves some duplication.
2. Add an integration test in `tests/`
3. Delete the work-around from `faq.md` in the docs.

AnnMarieW:
Great - I'm glad you like it. :smiley_cat:  I learned how to do that when I did the PR to add the Clipboard component to `dash-core-components`

I'll start working on your other requests.

tcbegley (approved):
Great, thank you!

Issue #730, opened 2021-10-09
V1 Themes Documentation includes reference to Bootstrap V4


glsdown:
Currently the v1 documentation states:

> _dash-bootstrap-components_ doesn't come with CSS included. This is to give you the freedom to use any Bootstrap v4 stylesheet of your choice, so you can achieve the look you want in your app.

As v1 is using Bootstrap 5, this should be updated.

AnnMarieW:
While someone is in the neighborhood, here are a couple more minor docs updates:

In the Icons section, we should make it really clear which version is used.  I've noticed that there are many changes between FontAwesome versions -- and the icons just don't appear if the syntax isn't correct, or certain icons from a different version are used. 

In the first paragraph in the layout section it says:
>The Bootstrap grid has twelve columns, and five responsive tiers (allowing you to specify different behaviours on different screen sizes, see below).  

But there are six responsive tiers now with the new XXL tier

glsdown:
@tcbegley I think that people wouldn't really know to do that though - I certainly wouldn't. Also, including in the docs makes it something people think about rather than spending ages on debugging why it isn't working.

tcbegley:
Icons page has snake_case `class_name` for HTML components which obviously won't work... 

https://dash-bootstrap-components.opensource.faculty.ai/docs/icons/

PR #747, opened 2021-10-21
Documentation improvements

glsdown:
Number of small improvements to documentation as detailed in #730 . This should close the remaining issues in there.

In addition, it includes small fixes for the `CardImg` and `Row` components.

- `CardImg`- the `top` and `bottom` props were not being correctly passed through to the underlying component
- `Row` - added support for `justify-content-evenly`


tcbegley (approved):
Thanks!

PR #759, opened 2021-11-03
Fixes default button color in docs

amarvin:
Docs for the `button` component (https://dash-bootstrap-components.opensource.faculty.ai/docs/components/button/) still show the default color as "secondary", although that changed to "primary" in Bootstrap5 and the migration guide (https://dash-bootstrap-components.opensource.faculty.ai/migration-guide/) shows that.


tcbegley (approved):
Good spot, thank you!

PR #812, opened 2022-02-08
Fix the link to Navbar docs in the migration guide

tg2648:
This fixes the link to Navbar docs in the migration guide which currently points to the ListGroup docs.


tcbegley (approved):
Great! Thank you.

tcbegley:
Change is live, thanks again!

Issue #813, opened 2022-02-08
Update fontawesome to v6


glsdown:
As FA v6 has now been released and is backwards compatible with 4 and 5, it would be great to add it.

PR #837, opened 2022-03-27
Upgrade Icon Packages

glsdown:
# Font Awesome

- Upgrades Font Awesome from 5.15.4 to 6.1.1 (per issue #813 )
- V6 is backwards compatible with V4 and V5, so this will not be a breaking change - the only major difference is that they have removed support for Internet Explorer. See details of what's changed in [their docs](https://fontawesome.com/docs/web/setup/upgrade/whats-changed).

# Bootstrap Icons

- Upgrades Bootstrap Icons from 1.5.0 to 1.8.1


tcbegley (approved):
Great, thanks!

Issue #834, opened 2022-03-26
Accordion items always open


VictorLFandino:
Hello

I wonder if its possible to implement an accordion where items do not close when others are opened. The idea is here:
[https://getbootstrap.com/docs/5.0/components/accordion/#always-open](https://getbootstrap.com/docs/5.0/components/accordion/#always-open)

Thank you very much

tcbegley:
Hi @VictorLFandino 

Great suggestion, was unaware of this functionality in Bootstrap. I can see it's supported in `react-bootstrap` (on which this library is built) so it should be possible to add.

PR #840, opened 2022-03-27
Add Accordion always_open property

glsdown:
As requested in issue #834 I've added support for `always_open` property of the Bootstrap 5 Accordion. This includes:

- JS tests
- Docs
- Doc tests

I have additionally bumped the react-bootstrap version.


glsdown:
@tcbegley - R and JL tests for the new accordion callback are failing (`test_accordion.py:check_always_open_callback_callbacks`). My guess is that the output format for the lists are different to in python (I'm not even sure if the way I am inserting them into the strings works for R or JL). As I don't have them installed, please can you take a look? Thanks!

glsdown:
@tcbegley - due to the fact that `always_open` does not work in R, I have removed this from the above and noted in the docs that it does not work with it. This will allow us to release for python and Julia, and revisit at a later date.

tcbegley (approved):

PR #866, opened 2022-06-10
Update migration-guide.md

3tilley:
Hello,

I was caught by this during a migration, so thought I would improve the docs for the next person


tcbegley:
Hey @3tilley

Thanks for the contribution! Could you explain what you mean by "Must be used in callbacks"?

3tilley:
Sorry that wasn't very clear. It might be that `.children` still works for setting the initial text, but is deprecated and `.label` is preferred in which case it's not a breaking change.

If you want to set the text in a Dash callback though, `Output("progress-id", "children")` will silently fail and render something odd. You have to update to `Output("progress-id", "label")` - which is why it was a breaking change for us

tcbegley:
Oh interesting, and `children` renders text ok if it's the initial value? Happy to merge in any case.

3tilley:
From what I could see, yes

PR #876, opened 2022-07-01
removed jumbotron since it's deprecated

Coding-with-Adam:


tcbegley (approved):
Thanks Adam! Changed the other examples and added a bit of styling in accordance with our docs. Not sure how this one slipped through!

PR #878, opened 2022-07-13
Forward undefined to target props instead of bool

schmijos:
_react-dom_ complains with a warning if it receives booleans for non-boolean DOM properties (see <https://github.com/facebook/react/blob/main/packages/react-dom/src/shared/ReactDOMUnknownPropertyHook.js#L186-L190>).

This passes `undefined` to the `target` props instead of `false to fix the warning.

CleanShot 2022-07-13 at 15 12 30@2x

### DIFF
--- a/src/components/dropdownmenu/DropdownMenuItem.js
+++ b/src/components/dropdownmenu/DropdownMenuItem.js
@@ -52,8 +52,6 @@ const DropdownMenuItem = props => {
   return (
     <RBDropdown.Item
       as={useLink ? Link : 'button'}
-      // don't pass href if disabled otherwise reactstrap renders item
-      // as link and the cursor becomes a pointer on hover
       href={useLink ? href : undefined}
       disabled={disabled}
       target={useLink ? target : undefined}
--- a/src/components/dropdownmenu/DropdownMenuItem.js
+++ b/src/components/dropdownmenu/DropdownMenuItem.js
@@ -54,9 +54,9 @@ const DropdownMenuItem = props => {
       as={useLink ? Link : 'button'}
       // don't pass href if disabled otherwise reactstrap renders item
       // as link and the cursor becomes a pointer on hover
-      href={disabled ? null : href}
+      href={useLink ? href : undefined}
       disabled={disabled}
-      target={useLink && target}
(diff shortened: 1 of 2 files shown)


src/components/listgroup/ListGroupItem.js
```diff
@@ -32,13 +33,14 @@ const ListGroupItem = props => {
...
   const isBootstrapColor = bootstrapColors.has(color);
-  const useLink = props.href && !disabled;
+  const useLink = href && !disabled;
   otherProps[useLink ? 'preOnClick' : 'onClick'] = incrementClicks;
 
   return (
     <RBListGroupItem
       as={useLink ? Link : 'li'}
-      target={useLink && target}
+      href={href}
```
schmijos, line 42: Has been forwarded with `otherProps` before.

src/components/dropdownmenu/DropdownMenuItem.js
```diff
@@ -54,9 +54,9 @@ const DropdownMenuItem = props => {
       as={useLink ? Link : 'button'}
       // don't pass href if disabled otherwise reactstrap renders item
       // as link and the cursor becomes a pointer on hover
-      href={disabled ? null : href}
+      href={useLink ? href : undefined}
```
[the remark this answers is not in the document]
tcbegley: Nice, this looks a lot cleaner and easier to understand what's going on (if we're not using a link, don't pass `href`).

I suggest we delete the comment which no longer makes that much sense, and is outdated in its references to reactstrap.

src/components/dropdownmenu/DropdownMenuItem.js
```diff
@@ -54,9 +54,9 @@ const DropdownMenuItem = props => {
       as={useLink ? Link : 'button'}
       // don't pass href if disabled otherwise reactstrap renders item
       // as link and the cursor becomes a pointer on hover
```
tcbegley, line 56:
```suggestion
```

tcbegley (approved):
Thanks for this @schmijos! 

Sorry for the slow response, this is a project I maintain in my spare time which has been lacking the last few weeks for various reasons. Appreciate the contribution!

Issue #811, opened 2022-02-07
Add shorthand support to mirror dash-core-components


tcbegley:
Dash 2.1.0 added shorthand support for `dcc.Dropdown`, `dcc.Checklist` and `dcc.RadioItems`. It would be good to support the same syntax with `dbc.Select`, `dbc.Checklist` and `dbc.RadioItems` respectively.

See [community post](https://community.plotly.com/t/dash-2-1-0-released/60548#h-2-rearranged-keyword-arguments-flexible-types-3) and [corresponding pull request](https://github.com/plotly/dash/pull/1763)

PR #894, opened 2022-10-15
Add shorthand support

glsdown:
As per issue #811 I've added shorthand support for Checklist, RadioItems and Select to bring it in line with dcc. I have added documentation and tests.

This means that all of these components take `options` and `value` as positional arguments, additionally `options` can take any of these forms:

### Original array of dictionaries:

```python
options=[
    {
      "label": "Label 1",
      "value": "value_1",
      ...
    },
    {
      "label": "Label 2",
      "value": "value_2",
      ...
    },
  ]
```

### List of values

```python
options = ["Option A", "Option B", "Option C"]
```

### Dictionary of value: label pairs

```python
options = {"A": "Option A", "B": "Option B", "C": "Option C"}
```

### TODO:

- [ ] The automatic docstring generation for the `options` parameter is quite confusing, and needs looking at, but I was unable to override it.
- [x] ~I don't have access to R or Julia, so have been unable to test the doc examples. These need checking and testing.~ No longer relevant as removed these examples due to issues with Dash in R and Julia


glsdown:
@tcbegley looks like the Julia and R doc snippet tests might need some tlc as they timed out at 6 hours

glsdown:
@tcbegley as we are no longer able to update the Julia and R package due to how it is bundled with Dash (as well as the fact it appears they are no longer supporting it), I have removed the Julia and R examples in the documentation, and added a note about the new functionality only being available in Python.

Issue #898, opened 2022-10-16
Update stylesheets


glsdown:
Bootstrap 5.2.2 is now available, along with potentially updated icons. The new styles sheets for these should be added.

Issue #844, opened 2022-04-02
Add newer features supported by React-Bootstrap


glsdown:
React Bootstrap have a number of features that we don't currently include. It would be good to add them:

#### Features
- [x] [Numbered List Groups](https://react-bootstrap.github.io/components/list-group/#numbered) (#895 )
- [x] [Stacks](https://react-bootstrap.netlify.app/layout/stack/#stacks) (#897 )
- [x] [Horizontal Collapse Animation](https://react-bootstrap.netlify.app/utilities/transitions/#horizontal) (#896 )
- [ ] [Custom Accordion Toggles](https://react-bootstrap.github.io/components/accordion/#custom-accordions)
- [x] [Placeholder](https://react-bootstrap.github.io/components/placeholder/) (#899)

#### Examples
- [ ] Combining the Navbar and Offcanvas components to make a [navigation system](https://react-bootstrap.github.io/components/navbar/#offcanvas).

tcbegley:
Nice. I don't know if the custom toggles will be possible as it seems it requires you pass in a function (which we can't replicate with Dash). Also not so sure about stacks. Feels like deviation from core Bootstrap functionality?

glsdown:
That might be why I originally left the custom toggles out of the Accordion.

Stacks are in Bootstrap v5: https://getbootstrap.com/docs/5.1/helpers/stacks/

tcbegley:
I stand corrected! Ok, let's put them in.

PR #899, opened 2022-10-16
Add new Placeholder component

glsdown:
As per #844 I have added the new Placeholder component. I have implemented it slightly differently to the React-Bootstrap component, using a button flag rather than a separate component.

- [ ] The 'glow' animation doesn't appear to be working but I wonder if this is in relation to #898


tcbegley (approved):
Very nice!

PR #940, opened 2023-02-21
Components as props in labels

AnnMarieW:
Added support for components in props for `Checklist, Checkbox, RadioButton, RadioItems`, and `Switch`

### DIFF
--- a/docs/components_page/components/input/components_in_labels.py
+++ b/docs/components_page/components/input/components_in_labels.py
@@ -1,5 +1,5 @@
 import dash_bootstrap_components as dbc
-from dash import Dash, html
+from dash import html
 
 flights = html.Div([html.I(className="bi bi-airplane pe-1"), "Flights"])
 car = html.Div([html.I(className="bi bi-car-front pe-1"), "Rental Car"])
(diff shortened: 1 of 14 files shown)


docs/components_page/components/input/components_in_labels.py
```diff
@@ -0,0 +1,31 @@
+from dash import Dash, html
+import dash_bootstrap_components as dbc
+
+flights = html.Div([html.Div(className="fa fa-plane pe-1"), "Flights"])
+car = html.Div([html.Div(className="fa fa-car pe-1"), "Rental Car"])
+hotel = html.Div([html.Div(className="fa fa-hotel pe-1"), "Hotel"])
```
tcbegley, line 6: These icons aren't showing up for me when I run the docs because Font Awesome stylesheet isn't actually linked.

We can add the link in `docs/templates/partials/head.html`, though I would actually have a preference for changing this example to use Bootstrap icons, since we're already linking the CSS anyway. So less for the reader of the docs to load when they visit.

tcbegley (changes requested):
Thanks for this @AnnMarieW 

I've made a few suggestions to get the tests working, but running your examples locally everything looks fine to me. This will be a great addition to the library!

tcbegley (approved):
Thanks for this!

I just updated the `noxfile.py` to format the tests and ran `nox -s format`. I also added the new bootstrap icons CDN link to `docs/templates/partials/head.html`, since that's where the docs pull the CSS from rather than from `dbc.icons`.

Hopefully we'll get all green CI and then I think this is good to go!

PR #946, opened 2023-02-28
Fixes #944

### LINKED ISSUE #944: Bug (dbc.Tab): active_label_style doesn't override label_style
- dash version: `#2.8.1`
- dash-bootstrap-components version: `#1.3.1`
- components affected by bug: dbc.Tab

### What is happening?

When setting a different value on the same css property in `label_style` and `active_label_style`
the value from `label_style` is always the one that is applied, 
Only properties from `active_label_style` that aren't set in `label_style` are applied.

Example : 
```py
tab_styles = {
    "label_style": {
        "borderRadius": "20px",         # always applied
        "width": "200px",               # always applied
        "textAlign": "center",          # always applied
        "color": "#676767",             # always applied even when active
        "backgroundColor": "#fff"       # always applied even when active
    },
    "active_label_style": {
        "backgroundColor": "#6699CC",   # not applied when active
        "color": "#fff",                # not applied when active
        "transform": "rotate(90deg)"    # applied when active
    },
}

layout = dbc.Tabs([
    dbc.Tab(dbc.Container(), label='Tab1', **tab_styles),
    dbc.Tab(dbc.Container(), label='Tab2', **tab_styles),
    dbc.Tab(dbc.Container(), label='Tab3', **tab_styles)
])
```

### What should be happening?

Normal behavior should be to have the style within `active_label_style` take priority over `label_style` for active tabs.
It works that way with `active_tab_style` and `tab_style`.

Nenrikido:


tcbegley (approved):
Perfect, thanks so much for your contribution!

PR #954, opened 2023-03-20
Deprecate R and Julia packages

tcbegley:
Unfortunately since upstream versions of Dash for R and Julia are poorly maintained and not widely used, it doesn't make much sense for us to keep maintaining these.

The packages will remain available on their respective branches and in the Julia package registry, but this PR removes the docs and code for building new versions.

TODO:
- [x] Replace HLJS bundle with one that doesn't have support for R and Julia syntax highlighting (less stuff to load when visiting the docs)
- [x] Make the example in `icons.md` look a bit prettier, i.e. perhaps wrap the alerts with icons + accompanying code in a card or similar. Could repurpose the old custom markdown parser for this.
- [ ] Simplify the release process now that we don't need to make R and Julia releases (should be possible to have everything triggered from release in GitHub UI + Actions to build + publish the Python / JS packages)


codercahol:
I am trying to using Dash with Julia and this would have been very useful for me!

Issue #970, opened 2023-09-04
Consider Supporting Hyperlinks in Carousel Images


mapix:
Before opening:

- [x] [Search for duplicate or closed issues](https://github.com/facultyai/dash-bootstrap-components/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [x] Read the [contributing guidelines](https://github.com/facultyai/dash-bootstrap-components/blob/main/.github/CONTRIBUTING.md)

Feature requests must include:

- [x]  As much detail as possible for what we should add and why it's important to
include it in Dash Bootstrap Components
   
In many display-type scenarios, it is not only necessary to present the content elements themselves, but also to guide users to further interaction with the content.

Usage example:
``` python
rec_section = dbc.Carousel(
    items=[
        {
            "key": _app["title"],
            "src": f"https://static01.dp.tech/image-proxy.mlops.dp.tech/v1/x340,s/{_app['cover']}",
            "header": _app["title"],
            "caption": _app["description"],
            "href": _app["href"],   # HERE what added
        }
        for _app in bohrium_apps
        if _app["recommended"]
    ],
    controls=True,
    className="carousel-fade",
    indicators=True,
    variant="dark",
    id="display",
    interval=2000,
    style={"height": "400px", "overflow": "hidden"},
)
```

- [x]  Relevant links to prior art, screenshots, or live demos whenever possible
https://app.bohrium.dp.tech/

I have simply implemented this feature, and I kindly ask the maintainer to evaluate whether this modification conforms to the project's design. If it is suitable, I will submit a Pull Request.

https://github.com/facultyai/dash-bootstrap-components/compare/main...pragmatic-dash:dash-bootstrap-components:feat/carousel-links?expand=1

tcbegley:
Hi @mapix

Thanks for the suggestion! I actually went ahead an implemented this independently as I didn't notice your final comment, sorry about that. You can see my changes in #971 and #973. You can also try it by installing the pre-release I made

```
pip install dash-bootstrap-components==1.5.0rc3
```

If you do try it with your app, I'd be keen to know how you get on.

There's a few differences with your implementation, including support for internal links, and making use of the `as` prop in `Carousel.Item` which means we don't have to create extra HTML elements that might interfere with Bootstrap's CSS.

I did however fail to implement the `target` attribute, if you want to submit a PR adding that to my changes I would accept such a change. lmk if you would be interested and think this could be a useful feature.

Here's an example I tested my changes with:

[...]

PR #975, opened 2023-09-05
support target for carousel image click

mapix:
As  https://github.com/facultyai/dash-bootstrap-components/issues/970

- [x] Add target property for Carousel item
- [x]  Add tests


tcbegley (approved):
LGTM, thanks for the contribution!

PR #991, opened 2023-12-29
updated carousel

AnnMarieW:
Currently the Carousel component is limited to displaying images. 

This PR is just a proof of concept to show that it's possible to have any components as slide content and captions.  The slides can be interactive too - callback for components in the slides work. 

Given that this PR includes breaking changes, I think it would be better to have a completely new Carousel component with `CarouselItem` and `CarouselCaption` components -- which is consistent with other components in the dbc library.


BSd3v:
This would be very helpful to create presentations at work. That way you can launch a presentation straight from the dash app.

PR #992, opened 2023-12-30
ci: Use GITHUB_OUTPUT envvar instead of set-output command

arunsathiya:
`save-state` and `set-output` commands used in GitHub Actions are deprecated and [GitHub recommends using environment files](https://github.blog/changelog/2023-07-24-github-actions-update-on-save-state-and-set-output-commands/).

This PR updates the usage of `set-output` to `$GITHUB_OUTPUT`

Instructions for envvar usage from GitHub docs:

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter


tcbegley:
Hey, thanks for proposing this and submitting the PR. Apologies it has taken me time to respond, I am quite stretched with other things atm and haven't had much time for this repo.

I like this change and would like to include it. I think this will need some additional changes though to work. See e.g. line 37 where we use the result of set-output. We need to change that to the environment variable.

Can we also use a more descriptive name such as DBC_VERSION or RELEASE_VERSION rather than GITHUB_OUTPUT?

arunsathiya:
No problem, thanks for responding! 

>See e.g. line 37 where we use the result of set-output. We need to change that to the environment variable.

`version` stores the value, so that reference/line is okay. 

>Can we also use a more descriptive name such as DBC_VERSION or RELEASE_VERSION rather than GITHUB_OUTPUT?

The variable/parameter (in the above example, `version`) is configurable `$GITHUB_OUTPUT` environment file is not.

tcbegley:
Oh! Got it, we're saving key-value pairs to a file specified by the variable, not saving the information to the variable itself 🤦‍♂️.

I clearly need to read things more carefully. Ok great, thanks for this then!

tcbegley (approved):

PR #999, opened 2024-02-01
Sanitize href props with xss vulnerability

AnnMarieW:
Sanitize html props that are vulnerable to xss vulnerability if user data is inserted.

Here's an example:

```

from dash import Dash, html

app=Dash(__name__)

app.layout = html.Div(
    dbc.NavLink("link",  href='javascript:alert(1)')
)

if __name__ == '__main__':
    app.run_server(debug=True)

```

Applied the same fix as  https://github.com/plotly/dash/pull/2732

The `href` prop is sanitized using the [braintree/sanitize-url](https://www.npmjs.com/package/@braintree/sanitize-url)

Content  with  vulnerabilities are replaced with "about:blank", and an error message is sent.

Dash 2.15 has a new prop:
> Add special key `_dash_error` to `setProps`, allowing component developers to send error without throwing in render. Usage `props.setProps({_dash_error: new Error("custom error")})`

I've included this, but it only sends the error message when using Dash 2.15.  Earlier Dash versions will not show an error in the console or the dev tools, but the prop is still sanitized.  

Would you like to bump the  `install_requires` to `dash>=2.15`  ?  I'm not sure how that would affect people that need to specify a dash version to get certain flask version. 

Before writing the tests would you like to take a quick look?  You can run `usage.py` to see all the components that are updated so far.

To Do:
- [ ] Carousel  
- [ ] Tests
- [ ] Missing any other components?

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,8 @@
     "ramda": "^0.27.1",
     "react": "^16.14.0",
     "react-bootstrap": "^2.2.2",
-    "react-dom": "^16.14.0"
+    "react-dom": "^16.14.0",
+    "@braintree/sanitize-url": "^7.0.0"
   },
   "jest": {
     "testEnvironment": "jsdom",
--- a/src/components/badge/Badge.js
+++ b/src/components/badge/Badge.js
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, {useEffect, useMemo} from 'react';
 import PropTypes from 'prop-types';
 import {omit} from 'ramda';
+import {sanitizeUrl} from '@braintree/sanitize-url';
 import RBBadge from 'react-bootstrap/Badge';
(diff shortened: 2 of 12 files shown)


tcbegley (commented):
Thanks for this!

- Can you add `package-lock.json`. The mismatched dependencies in `package.json` is causing CI to fail
- Requiring Dash 2.15 feels aggressive. If things still work in older versions but we're missing a dev feature like better error messages then that's ok
- In addition to unit tests, can we write integration test that checks the error is raised?
- the code is a bit repetitive, do you think we could maybe refactor into a custom hook in `src/private`? If it works we could even upstream into `@plotly/dash-component-plugins` then just import it!

AnnMarieW:
Closed in favor of #1000

PR #1000, opened 2024-02-03
Sanitize href props with xss vulnerability V2

AnnMarieW:
Sanitize html props that are vulnerable to xss vulnerability if user data is inserted.

This is the new version based on comments in #999 

Let's keep this as a draft until https://github.com/plotly/dash/issues/2743 is resolved.


src/components/nav/NavLink.js
```diff
@@ -1,9 +1,11 @@
-import React, {useEffect, useState} from 'react';
+import React, {useEffect, useState, useMemo} from 'react';
```
tcbegley, line 1: `useMemo` not used in this file afaict

tcbegley (commented):
Looking good so far.

1000th issue / PR! Cool milestone 😄

src/components/nav/NavLink.js
AnnMarieW: Right - I didn't do `npm run format` yet.  Still lots of files to change :-)

tcbegley (approved):
Great stuff, thanks @AnnMarieW!

I made a few small edits:
* Used `pytest.mark.parameterize` inside the integration tests, I think it makes it a little easier to see what is going on in the test + that the same check is being performed for multiple components.
* Some small changes to prop destructuring, e.g. where we pull out `setProps` we no longer need to call `omit` and so on.

PR #1001, opened 2024-02-06
add enforceFocus property to the modal

arthur-FD:
Hello,

Adding the enforceFocus property to the modal.
It is set by default to true, and cannot be change otherwise.
It is usefull to be able to set it to false when you have other elements you need to be able to interact with when the modal is activated - in our usecase, its a feedback widget.

Arthur


tcbegley (approved):
Thanks for the contribution!

tcbegley:
This is available in the latest beta

```
pip install dash-bootstrap-components==1.6.0b1
```

PR #1007, opened 2024-04-14
[RFC] TooltipWrapper component

tcbegley:
This PR implements an experimental `TooltipWrapper` component which attaches a tooltip to the components that it wraps rather than using IDs to match to elements in the DOM.

Example usage would be as follows.

```python
from dash import Dash, html

app = Dash(external_stylesheets=[dbc.themes.BOOTSTRAP])

app.layout = dbc.Container(
    [
        html.P("The button below has a tooltip"),
        dbc.TooltipWrapper(dbc.Button("Hover over me!"), text="test tooltip"),
    ],
    className="p-5",
)

app.run_server(debug=False)
```

I'm not really sure about this. It's kind of neat, but I'm not sure it really solves any major problems that the original implementation has, and I generally think it's preferable for there to not be multiple different ways of accomplishing the same thing.

### DIFF
--- a/src/components/tooltip/TooltipWrapper.js
+++ b/src/components/tooltip/TooltipWrapper.js
@@ -1,0 +1,127 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {OverlayTrigger, Tooltip} from 'react-bootstrap';
+
+/**
+ * An alternative way of creating tooltips.
+ */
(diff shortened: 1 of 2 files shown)


tcbegley:
Closing as I don't think this is a significant improvement on the existing ways of doing things, and in general I think it's better to have only one recommended way of doing things.

Issue #1119, opened 2025-04-14
Tab component being enabled via callback not applied correctly


FRosner:
Before opening:

- [Search for duplicate or closed issues](https://github.com/facultyai/dash-bootstrap-components/issues?utf8=%E2%9C%93&q=is%3Aissue)
- Read the [contributing guidelines](https://github.com/facultyai/dash-bootstrap-components/blob/main/.github/CONTRIBUTING.md)

Please fill out the below information as much as possible.

- dash version: `3.0.2`
- dash-bootstrap-components version: `2.0.1`
- components affected by bug:

### What is happening?

When updating the `disabled` property of `Tab` components within a `Tabs` component via a callback, the change does not get "rendered" in the UI properly. It does get applied once the user switches to another tab, though. See the attached recording:

https://github.com/user-attachments/assets/912787d9-d012-4e77-8622-eb41201e9782

### What should be happening?

All tabs should be enabled after the callback ran, which runs immediately after the tabs are rendered. This works as expected in the previous version we used (1.6.0).

### Code

```python
import dash
from dash import html, dcc, Input, Output, State, callback
import dash_bootstrap_components as dbc

app = dash.Dash(__name__, external_stylesheets=[dbc.themes.BOOTSTRAP])

app.layout = html.Div([
    html.H1("Dash Bootstrap Tabs Example - Disabled Tab"),
    html.Hr(),

    dbc.Tabs(
        id="tabs",
        children=[
            dbc.Tab(label="Tab 1", tab_id="tab-1", id="tab-1-content", children=[
                html.Div([
                    html.H3("Content of Tab 1"),
                    html.P("This tab is enabled by default.")
                ], className="p-4")
            ]),
            dbc.Tab(label="Tab 2", tab_id="tab-2", id="tab-2-content", children=[
                html.Div([
                    html.H3("Content of Tab 2"),
                    html.P("This tab was initially enabled")
                ], className="p-4")
            ]),
            dbc.Tab(label="Tab 3", tab_id="tab-3", id="tab-3-content", disabled=True, children=[
                html.Div([
                    html.H3("Content of Tab 3"),
                    html.P("This tab was initially disabled but has been enabled via callback.")
                ], className="p-4")
            ]),

        ],
        active_tab="tab-1",
    ),
])

@callback(
    Output("tab-3-content", "disabled"),
    Input("tabs", "children"),
)
def enable_tab(_):
    return False

if __name__ == "__main__":
    app.run(debug=True)
[...]

PR #1120, opened 2025-04-15
#1119 Make sure tab updates are rendered correctly

FRosner:
# Summary

This PR sets `dashChildrenUpdate` to `true` on the tabs component as advised by https://github.com/plotly/dash/pull/3268 to address a regression introduced in Dash 3 that would prevent tabs from getting rendered properly after updating them via a callback.

# References

- Fixes https://github.com/facultyai/dash-bootstrap-components/issues/1119

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ authors = [
 maintainers = [{ name = "Tom Begley", email = "tomcbegley@gmail.com" }]
 requires-python = ">=3.9"
 dependencies = [
-    "dash>=3.0.0",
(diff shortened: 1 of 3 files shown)


tcbegley (approved):
Great, thanks for this! Let me just have a think about whether there are other components where this change would be relevant and then I will merge.

tcbegley:
We're all good, I was wondering if `Accordion` and `AccordionItem` might need a similar update, but they work differently to `Tabs` and `Tab` so we're all good.

